PR #590, opened 2020-11-10
Ignore all the build fails and continue the process with warning

ThisaruGuruge:
$ subject

### DIFF
--- a/release/src/publish/publish.bal
+++ b/release/src/publish/publish.bal
@@ -148,7 +148,7 @@ function isWorkflowRunSuccess(map<json> payload, commons:Module module) returns 
     map<json> workflowRun = getWorkflowJsonObject(payload);
     string conclusion = workflowRun.conclusion.toString();
     if (conclusion != CONCLUSION_SUCCSESS) {
-        string message = "Module " + module.name + " build did not successfully completed.";
+        string message = "Module '" + module.name + "' build has not completed successfully.";
         log:printWarn(message + " Conclusion: " + conclusion);
     }
     return true;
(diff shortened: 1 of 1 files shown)


release/src/publish/publish.bal
```diff
@@ -147,18 +147,11 @@ function isWorkflowCompleted(map<json> payload) returns boolean {
...
-    if (conclusion == CONCLUSION_SUCCSESS) {
-        return true;
-    }
-    string message = "Module " + module.name + " build did not successfully completed.";
-    error e = error("Unsuccessfull", message = "Workflow run conclusion: " + conclusion);
-    if (module.dependentModules.length() > 0) {
-        commons:logAndPanicError(message, e);
-    } else {
+    if (conclusion != CONCLUSION_SUCCSESS) {
+        string message = "Module " + module.name + " build did not successfully completed.";
```
niveathika, line 151:
```suggestion
        string message = "Module '" + module.name + "' build has not completed successfully.";
```

release/src/publish/publish.bal
```diff
@@ -147,18 +147,11 @@ function isWorkflowCompleted(map<json> payload) returns boolean {
...
-    if (module.dependentModules.length() > 0) {
-        commons:logAndPanicError(message, e);
-    } else {
+    if (conclusion != CONCLUSION_SUCCSESS) {
+        string message = "Module '" + module.name + "' build has not completed successfully.";
         log:printWarn(message + " Conclusion: " + conclusion);
-        return true; // Returning true, since we don't want these builds succeed to continue to the next level.
     }
-    return false;
+    return true;
```
niveathika, line 154: It seems now it returns true always? In that case, we can change function, can't we?

ThisaruGuruge: Um, it will complicate the change. We do return this function in another place directly. Let's keep it as it is for now. We may change it later.

ThisaruGuruge: Fixed with 7df6c91d58fa6528b52be74670983d028dfef039.

niveathika (approved):

PR #600, opened 2020-11-12
Logs added to main.py

TharindaDilshan:
Logs to determine the workflow status

### DIFF
--- a/release/src/module_dependencies/main.py
+++ b/release/src/module_dependencies/main.py
@@ -18,23 +18,35 @@ def main():
 
 # Sorts the ballerina standard library module list in ascending order
 def sortModuleNameList():
-    with open('./release/resources/module_list.json') as f:
-          nameList = json.load(f)
+    try:
+        with open('./release/resources/module_list.json') as f:
+            nameList = json.load(f)
+    except:
+        print('Failed to read module_list.json')
+        sys.exit()
 
     nameList['modules'].sort()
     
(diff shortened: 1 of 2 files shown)


ThisaruGuruge:
How can we identify any error scenarios? IMO, we need the logs the most when something is wrong. Say an HTTP call fails, or file read error, etc. Shall we try to identify those error scenarios and log them?

TharindaDilshan:
> How can we identify any error scenarios? IMO, we need the logs the most when something is wrong. Say an HTTP call fails, or file read error, etc. Shall we try to identify those error scenarios and log them?

I updated the code to log exceptions

ThisaruGuruge (approved):

release/src/module_dependencies/main.py
```diff
@@ -2,33 +2,59 @@
...
+            jsonFile.truncate()
+    except:
+        print('Failed to write to file module_list.json')
+        sys.exit()
         
     return nameList['modules'] 
 
+# Returns the file in the given url
+# Retry decorator will retry the function 3 times, doubling the backoff delay if URLError is raised 
+@retry(urllib.error.URLError, tries=3, delay=2, backoff=2)
```
ThisaruGuruge, line 45: What is the unit of this delay? 

Shall we define proper constants for the `tries`, `delay`, and the `backoff` values and use them here?

ThisaruGuruge (approved):

PR #605, opened 2020-11-17
Create PR for dependency updates

TharindaDilshan:
Creating a PR instead of directly committing dependency updates

### DIFF
--- a/.github/workflows/dependency_graph.yml
+++ b/.github/workflows/dependency_graph.yml
@@ -12,6 +12,21 @@ jobs:
         runs-on: ubuntu-latest
         steps:
             -   uses: actions/checkout@v2
+
+            -   name : Configure GitHub and create branch
+                env:
+                    GITHUB_TOKEN: ${{ secrets.BALLERINA_BOT_TOKEN }}
+                run: |
+                    git config user.name ${{ secrets.BALLERINA_BOT_USERNAME }}
+                    git config user.email ${{ secrets.BALLERINA_BOT_EMAIL }}
+                    echo "BRANCH_EXISTS=0" >> $GITHUB_ENV
+                    if git ls-remote --heads origin dependency-graph | grep dependency-graph; then
+                        echo "BRANCH_EXISTS=1" >> $GITHUB_ENV
+                        git fetch
+                        git checkout dependency-graph
+                    else
+                        git checkout -b dependency-graph
+                    fi
         
             -   name: Setup Python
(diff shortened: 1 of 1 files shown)


.github/workflows/dependency_graph.yml
```diff
@@ -12,6 +12,21 @@ jobs:
...
         steps:
             -   uses: actions/checkout@v2
+
+            -   name : Configure GitHub and create branch
+                env:
+                    GITHUB_TOKEN: ${{ secrets.BALLERINA_BOT_TOKEN }}
+                run: |
+                    git config user.name ${{ secrets.BALLERINA_BOT_USERNAME }}
+                    git config user.email ${{ secrets.BALLERINA_BOT_EMAIL }}
+                    echo "BRANCH_EXISTS=0" >> $GITHUB_ENV
```
ThisaruGuruge, line 22: Can't we directly set this as an environment variable similar to line no. 18?

.github/workflows/dependency_graph.yml
```diff
@@ -28,21 +43,21 @@ jobs:
...
+                        git remote set-url origin https://${{ secrets.BALLERINA_BOT_USERNAME }}:${{ secrets.BALLERINA_BOT_TOKEN }}@github.com/ballerina-platform/ballerina-standard-library.git
+                        git add .
+                        git commit -m "stdlib_modules.json updated" || echo "No changes to commit"
+                        git push origin dependency-graph
+                        if hub pr list | grep 'StdLib Module Dependencies Updated'; then    
+                            echo "PR already exists"
+                        else
+                            echo "Creating PR"
+                            curl -fsSL https://github.com/github/hub/raw/master/script/get | bash -s 2.14.1
+                            bin/hub pull-request -m "StdLib Module Dependencies Updated"
```
ThisaruGuruge, line 61:
```suggestion
                            bin/hub pull-request -m "[Automated] Update stdlib module dependencies"
```

.github/workflows/dependency_graph.yml
TharindaDilshan: The value is being updated dynamically.

ThisaruGuruge: Still, we need to define it first right? This line is just defining the `BRANCH_EXISTS` variable with the initial value of `0`.

TharindaDilshan: Updated

ThisaruGuruge (approved):

PR #610, opened 2020-11-18
Regex to identify module version

TharindaDilshan:


release/src/module_dependencies/main.py
```diff
@@ -86,7 +86,7 @@ def getVersion(balModule):
     version = ''
     for line in data:
         processedLine = line.decode("utf-8")
-        if 'version' in processedLine:
+        if re.match('version=', processedLine):
```
ThisaruGuruge, line 89: This would match anything with `version=` right?

What if there's an entry with `someversion=x.x.x`?

```suggestion
        if re.match('^\s+version=', processedLine):
```

TharindaDilshan: This only identifies strings starting with `version=` 
`someversion=` will not be matched

ThisaruGuruge: It will. That's the way Regex works right? `someVersion=` will not be matched, but `someversion=` will. Isn't it?

TharindaDilshan: Python [match()](https://docs.python.org/3/library/re.html#re.Pattern.match) function only matches if the string starts with the given pattern

release/src/module_dependencies/main.py
```diff
@@ -159,8 +162,6 @@ def calculateLevels(moduleNameList, moduleDetailsJSON):
 
 # Updates the stdlib_modules.JSON file with dependents of each standard library module
 def updateJSONFile(updatedJSON):
-    updatedJSON['modules'].sort(key=lambda s: s['level'])
```
ThisaruGuruge, line 162: Are we removing the sorting part?

release/src/module_dependencies/main.py
```diff
@@ -197,4 +198,55 @@ def getImmediateDependents(moduleNameList, moduleDetailsJSON):
...
+    for line in readMeFile:
+        processedLine = line.decode("utf-8")
+        updatedReadMeFile += processedLine
+        if "|:---:|:---:|:---:|:---:|:---:|:---:|:---:|" in processedLine:
+            break
+
+    # Modules in levels 0 and 1 are categorized under level 1
+    # A single row in the table is created for each module in the module list    
+    levelColumn = 1
+    tempLevel = 0
```
ThisaruGuruge, line 220: A better name would be `currentLevel`

release/src/module_dependencies/main.py
```diff
@@ -197,4 +198,55 @@ def getImmediateDependents(moduleNameList, moduleDetailsJSON):
...
+    for module in moduleDetailsJSON['modules']:
+        if tempLevel != module['level'] and module['level'] != 1:
```
ThisaruGuruge, line 222: Can't this be simplified as this:

```
if module['level'] > levelColumn:
    levelColumn = module['level']
```

release/src/module_dependencies/main.py
```diff
@@ -197,4 +198,55 @@ def getImmediateDependents(moduleNameList, moduleDetailsJSON):
...
+            levelColumn = module['level']
+            tempLevel = module['level']
+
+        row = ("|" + str(levelColumn) + "|[" + module['name'].split('-')[-1] + "](https://github.com/ballerina-platform/"
```
ThisaruGuruge, line 226: We are going to print the level in each line here, right? That's not the way it is now.

We can overcome this by using a variable here and setting it on the condition above (Whether we are at a new level or not).

release/src/module_dependencies/main.py
```diff
@@ -197,4 +198,55 @@ def getImmediateDependents(moduleNameList, moduleDetailsJSON):
...
+    # A single row in the table is created for each module in the module list    
+    levelColumn = 1
+    tempLevel = 0
+    for module in moduleDetailsJSON['modules']:
+        if tempLevel != module['level'] and module['level'] != 1:
+            levelColumn = module['level']
+            tempLevel = module['level']
+
+        row = ("|" + str(levelColumn) + "|[" + module['name'].split('-')[-1] + "](https://github.com/ballerina-platform/" 
+        + module['name'] + ")| [![GitHub Release](https://img.shields.io/github/release/ballerina-platform/"
```
ThisaruGuruge, line 227: It's better if we can use constants for these links and texts. That way it will be easier to read and to make any changes in the future.

release/src/module_dependencies/main.py
TharindaDilshan: I've moved it to the main method as I'm using it to update the README file as well

release/src/module_dependencies/main.py
TharindaDilshan: I will update

release/src/module_dependencies/main.py
TharindaDilshan: Yes this is better. can get rid of the other variable

release/src/module_dependencies/main.py
TharindaDilshan: No this doesn't update the level on each row. I'm setting the level to an empty string at the end of the loop

release/src/module_dependencies/main.py
TharindaDilshan: Will do.

release/src/module_dependencies/main.py
```diff
@@ -197,4 +200,61 @@ def getImmediateDependents(moduleNameList, moduleDetailsJSON):
...
+    currentLevel = 0
+    for module in moduleDetailsJSON['modules']:
+        if module['level'] > currentLevel:
+            levelColumn = module['level']
+            currentLevel = module['level']
+
+        row = ("|" + str(levelColumn) + "|" + 
+        "[" + module['name'].split('-')[-1] + "](" + BALLERINA_ORG_URL + module['name'] + ")| " + 
+
+        "[![GitHub Release](" + GITHUB_BADGE_URL + "release/ballerina-platform/" + module['name'] + ".svg?label=)]" +
```
ThisaruGuruge, line 231: Can't we use a separate variable for `ballerina-platform`? I see it is used in many places.

TharindaDilshan: Done

ThisaruGuruge (approved):

PR #613, opened 2020-11-19
Remove filepath module from build pipeline

ThishaniLucas:
$subject

### DIFF
--- a/release/resources/module_list.json
+++ b/release/resources/module_list.json
@@ -7,7 +7,6 @@
         "module-ballerina-email",
         "module-ballerina-encoding",
         "module-ballerina-file",
-        "module-ballerina-filepath",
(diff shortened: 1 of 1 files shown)


niveathika:
@ThishaniLucas Shall we update the dashboard table too?

ThishaniLucas:
Seems like it's going to be automated in a couple of days. @ThisaruGuruge

ThisaruGuruge:
> Seems like it's going to be automated in a couple of days. Already discussed with @ThisaruGuruge

Yes, it is already in the review state. Most probably we can have it by end of this week. Shall we keep this as it is until then? We might use this as a test for that functionality as well.

ThisaruGuruge (approved):
LGTM

Issue #714, opened 2020-12-09
Automatically Update stdlib Versions in Ballerina Distribution


ThisaruGuruge:
After a stdlib version update, [we already have a workflow](https://github.com/ballerina-platform/ballerina-standard-library/actions?query=workflow%3A%22Stdlib+Module+Dependencies%22) to update the versions in stdlib repo. 

But the issue is the Ballerina distribution does not get updated to these versions. It's better if we can extend the existing workflow, or create a new workflow to send an automated PR to update the versions in the Ballerina distribution repo.

PR #766, opened 2020-12-18
Update stdlib Versions in Ballerina Distribution

TharindaDilshan:
Fixes: #714

### DIFF
--- a/release/src/version_update/main.py
+++ b/release/src/version_update/main.py
@@ -1,0 +1,4 @@
+import urllib.request
+import json
+import sys
+from retry import retry
--- a/release/src/version_update/main.py
+++ b/release/src/version_update/main.py
@@ -1,4 +1,153 @@
 import urllib.request
 import json
 import sys
+import os
 from retry import retry
+
+HTTP_REQUEST_RETRIES = 3
+HTTP_REQUEST_DELAY_IN_SECONDS = 2
+HTTP_REQUEST_DELAY_MULTIPLIER = 2
+packageUser =  os.environ["packageUser"]
+packagePAT = os.environ["packagePAT"]
+packageEmail =  os.environ["packageEmail"]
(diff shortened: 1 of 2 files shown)


release/src/version_update/main.py
```diff
@@ -0,0 +1,173 @@
...
+            moduleList = json.load(f)
+    except:
+        print('Failed to read stdlib_modules.json')
+        sys.exit()
+
+    return moduleList['modules']
+
+# Fetch ballerina-distribution repository with GitHub credentials
+def configureGithubRepository():
+    g = Github(packagePAT)
```
ThisaruGuruge, line 41: Shall we use a proper name for this?
Please check the other places as well.

release/src/version_update/main.py
```diff
@@ -0,0 +1,173 @@
...
+
+    data = file.decoded_content.decode("utf-8")
+
+    return data
+
+# Update stdlib module versions in the gradle.properties file with module details fetched from stdlib_modules.json
+def updatePropertiesFile(data, modules):
+    modifiedData = ''
+    updatedModules = []
+    pointer = ''
```
ThisaruGuruge, line 65: What does this `pointer` do? The name is ambiguous.

TharindaDilshan: It points to a specific line in `gradle.properties`

ThisaruGuruge: Then maybe `currentLine` would be a preferable name. Shall we rename it?

TharindaDilshan: Updated

release/src/version_update/main.py
ThisaruGuruge: I meant the variable `g` here.

ThisaruGuruge (approved):

PR #783, opened 2020-12-22
Add ballerina stdlib dependabot workflow

TharindaDilshan:
## Purpose

Add Ballerina Standard Library Dependabot workflow to automate the process of discovering the latest versions of the Standard Library modules and updating their dependent modules to the latest version.


release/src/stdlib_dependabot/file_update/main.py
```diff
@@ -0,0 +1,226 @@
+import urllib.request
```
ThisaruGuruge, line 1: Let's move this out to a separate directory at the top level. (Name maybe something like `dependabot`). 
This doesn't feel like belonging to the `release` directory.

release/src/stdlib_dependabot/file_update/main.py
```diff
@@ -0,0 +1,226 @@
...
+    updatedModuleDetails = initializeModuleDetails(modulesWithVersionUpdates)
+    updatedModuleDetails = getImmediateDependents(stdlibModules, updatedModuleDetails)
+    updateFiles(updatedModuleDetails)
+
+# Fetch the updated module versions string from the environment and append it to an array
+def preprocessString():
+    try:
+        # String is in the form -> '"module": "version" "module": "version"'
+        updatedModules = os.environ["modules"]
+        updatedModules = updatedModules.split()
```
ThisaruGuruge, line 29:
```suggestion
        updatedModules = os.environ["modules"].split()
```

release/src/stdlib_dependabot/file_update/main.py
```diff
@@ -0,0 +1,226 @@
...
+        sys.exit()
+
+# Fetch list of stdlib modules from stdlib_latest_versions.json 
+def getModuleListFromFile():
+    try:
+        with open('./release/resources/stdlib_latest_versions.json') as f:
+            moduleList = json.load(f)
+    except:
+        print('Failed to read stdlib_latest_versions.json')
+        sys.exit()
```
ThisaruGuruge, line 49: Shall we add an exit code here? This must be `!=0` right?

Check the other places as well.

release/src/stdlib_dependabot/version_update/main.py
```diff
@@ -0,0 +1,75 @@
+import urllib.request
```
ThisaruGuruge, line 1: Instead of having `main.py` in two separate directories, can't we have `version_update.py` and `file_update.py` at the same directory?

release/src/stdlib_dependabot/file_update/main.py
```diff
@@ -0,0 +1,226 @@
...
+                    data, 
+                    contents.sha, 
+                    branch="dependabot/" + module, 
+                    author=author)
+
+# Create a PR from the branch created
+def createPullRequest(repo, currentVersion, module, latestVersion):
+    pulls = repo.get_pulls(state='open', head='dependabot/' + module)
+
+    PRExists = 0
```
ThisaruGuruge, line 202:
```suggestion
    prExists = 0
```

The practice we use is if there is an abbreviation in a variable name, we consider the abbreviation as another word. If it is in the middle of a variable name, we only capitalize the first letter.

(In python this is completely different, I agree. But since we did not use the _python way_ for any of this, I think it's better to follow what we follow in other languages.)

release/src/stdlib_dependabot/file_update/main.py
TharindaDilshan: Updated

.github/workflows/stdlib_dependabot.yml
```diff
@@ -0,0 +1,70 @@
+name: Ballerina Stdlib Dependabot
+
+on: 
+    workflow_dispatch:
+    schedule:
+        - cron: '30 18 * * *'
+jobs:
+    dependabot:
+        name: Stdlib Dependabot
```
ThisaruGuruge, line 9: Shall we add the owner check here similar to what we did in the other workflows?

TharindaDilshan: Will do

ThisaruGuruge (approved):

PR #884, opened 2021-01-26
Fix JavaArrays KeyError

TharindaDilshan:
Change `java.arrays` name in gradle.properties from `stdlibJarraysVersion` to `stdlibJavaArraysVersion`

### DIFF
--- a/release/src/version_update/main.py
+++ b/release/src/version_update/main.py
@@ -13,6 +13,9 @@ packageUser =  os.environ["packageUser"]
 packagePAT = os.environ["packagePAT"]
 packageEmail =  os.environ["packageEmail"]
 organization = 'ballerina-platform'
+standardLibrary = 'stdlib'
+javaArraysModuleName = 'stdlibJavaArraysVersion'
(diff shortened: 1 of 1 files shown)


release/src/version_update/main.py
```diff
@@ -111,8 +111,8 @@ def updatePropertiesFile(data, modules, currentVersions):
         latestVersion = module['version']
 
         if moduleName == 'java.arrays':
-            version = compareVersion(latestVersion, currentVersions['stdlibJarraysVersion'])
-            line = "stdlibJarraysVersion=" + version + "\n"
+            version = compareVersion(latestVersion, currentVersions['stdlibJavaArraysVersion'])
```
ThisaruGuruge, line 114: Since we are using the same string in different places, it's better if we use a constant for this string value. That way, it will be easier to make changes to these.

TharindaDilshan: Updated

ThisaruGuruge (approved):

PR #914, opened 2021-01-29
Update stdlib dependabot to create single PR for all dependent modules

TharindaDilshan:
With this update, Ballerina Stdlib Dependabot will only create a single PR for all the dependent modules in each Stdlib module repository.

### DIFF
--- a/dependabot/update_repositories.py
+++ b/dependabot/update_repositories.py
@@ -162,6 +162,15 @@ def updatePropertiesFile(data, module, latestVersion):
                     modifiedLine = 'stdlib' + module.split('-')[-1].capitalize() + 'Version=' + latestVersion + "\n"
                 modifiedData += modifiedLine
                 commitFlag = True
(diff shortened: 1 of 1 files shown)


dependabot/update_repositories.py
```diff
@@ -184,49 +194,45 @@ def commitChanges(data, currentVersion, repo, module, latestVersion):
...
         try:
-            repo.create_pull(title="Bump " + module + " from " + currentVersion + " to " + latestVersion, 
+            repo.create_pull(title="Bump stdlib module versions", 
                             body='$subject', 
-                            head="dependabot/" + module, 
+                            head=dependabotBranchName, 
                             base="main")
         except GithubException:
-            repo.create_pull(title="Bump " + module + " from " + currentVersion + " to " + latestVersion, 
+            repo.create_pull(title="Bump stdlib module versions",
```
ThisaruGuruge, line 233: Title should have `[Automated]` prefix

TharindaDilshan: Updated

ThisaruGuruge (approved):

PR #966, opened 2021-02-11
Disable already released module releases

ThisaruGuruge:
$Subject

### DIFF
--- a/release/resources/stdlib_modules.json
+++ b/release/resources/stdlib_modules.json
@@ -4,7 +4,7 @@
             "name": "module-ballerina-encoding",
             "version": "1.1.0-alpha2-SNAPSHOT",
             "level": 1,
-            "release": true,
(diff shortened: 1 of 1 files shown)


BuddhiWathsala:
It seems the time module also successfully released, we need to disable that too.

https://github.com/ballerina-platform/ballerina-standard-library/runs/1881492007?check_suite_focus=true

ldclakmal:
> It seems the time module also successfully released, we need to disable that too.
> 
> https://github.com/ballerina-platform/ballerina-standard-library/runs/1881492007?check_suite_focus=true

Seems like `time` is released with older version pattern. Not with `-alpha2`. Hence we may be need to release this. Right?

BuddhiWathsala:
Noted and revert the commit.

ldclakmal (approved):

PR #1075, opened 2021-03-10
Update stdlib dependabot logs

TharindaDilshan:


dependabot/update_repositories.py
```diff
@@ -116,25 +116,26 @@ def get_dependencies(module_name):
...
                     commit_changes(updated_data, current_version, repo, module['name'], module['version'])
                     create_pull_request(repo, current_version, module['name'], module['version'])
-                    print("Bump " + module['name'] + " version from " + current_version + " to " + module['version'])
+                    print("Bump version in " + dependent + " from " + current_version)
                     time.sleep(30)
                 except:
                     continue
             else:
-                print(module['name'] + " version is already the lastest version " + current_version)
+                if current_version:
```
ThisaruGuruge, line 137: What does this condition do?

TharindaDilshan: If the current version in not set that means the version is not defined in the properties file. It is handled when updating the properties file. If the current version check is not done here, it will create two logs.

ThisaruGuruge (approved):

PR #1328, opened 2021-05-05
Authenticate api calls and change color to bright green

niveathika:


release/src/module_dependencies/update_version_data.py
```diff
@@ -180,37 +196,40 @@ def calculate_levels(module_name_list, module_details_json):
...
+            'version': version,
             'level': 0,
             'default_branch': default_branch,
-            'release': True, 
-            'dependents': [] })
-        
+            'release': True,
+            'dependents': []})
+
         time.sleep(10)
```
ThisaruGuruge, line 228: We can reduce the sleep time here now.

niveathika: Fixed in a9e6442

ThisaruGuruge (approved):

PR #1454, opened 2021-06-13
Improve dependency graph generator

niveathika:
$subject since it is recommended to only use platform level URL to identify packages repositories. Because of this, we cannot identify dependencies based on the repository URL. 

The script is improved to parse the gradle.properties file to identify the version key of each modules' dependencies. 

Default version_key is stdlib + Capitalised artifact Name + Version
This can be overridden in the modules.json.

Improvement because of https://github.com/ballerina-platform/ballerina-release/issues/547

### DIFF
--- a/release/resources/module_list.json
+++ b/release/resources/module_list.json
@@ -1,41 +1,118 @@
 {
     "modules": [
-        "module-ballerina-auth",
-        "module-ballerina-cache",
-        "module-ballerina-crypto",
-        "module-ballerina-email",
-        "module-ballerina-file",
-        "module-ballerina-ftp",
-        "module-ballerina-graphql",
-        "module-ballerina-grpc",
-        "module-ballerina-http",
-        "module-ballerina-io",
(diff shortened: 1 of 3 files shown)


ThisaruGuruge (approved):

ThisaruGuruge:
We should ask the repo owners to clean up the gradle.properties file after this. I think there are places where gradle.properties file has versions that aren't used in the build, and some of the gradle.properties files may not follow the naming convention too. (Especially we have to check the cases like OAuth2).

niveathika:
> We should ask the repo owners to clean up the gradle.properties file after this. I think there are places where gradle.properties file has versions that aren't used in the build, and some of the gradle.properties files may not follow the naming convention too. (Especially we have to check the cases like OAuth2).

@ThisaruGuruge Tracked it here, https://github.com/ballerina-platform/ballerina-release/issues/547#issuecomment-860365346

PR #2668, opened 2022-02-10
Update load-test badge generation script

ayeshLK:
This change is done to use the same load-test badge for WebSub and WebSubHub modules.

### DIFF
--- a/release/src/dependencies/utils.py
+++ b/release/src/dependencies/utils.py
@@ -20,7 +20,7 @@ def get_dashboard_row(module, level):
     codecov_badge = get_codecov_badge(module_name, default_branch)
     bugs_badge = get_bugs_badge(module_name)
     pull_requests_badge = get_pull_requests_badge(module_name)
-    load_tests_badge = get_load_tests_badge(module_name)
+    load_tests_badge = get_load_tests_badge(module_name, default_branch)
 
     return f'|{level}|{repo_link}|{release_badge}|{build_status_badge}|{trivy_badge}|{codecov_badge}|{bugs_badge}|{pull_requests_badge}|{load_tests_badge}|\n'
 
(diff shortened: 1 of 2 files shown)


release/src/dependencies/utils.py
```diff
@@ -94,10 +94,14 @@ def get_bug_query(module_name):
     encoded_filter = urllib.parse.quote_plus(issue_filter)
     return f'{encoded_filter}&label=&color={label_colour}'
 
-def get_load_tests_badge(module_name):
+def get_load_tests_badge(module_name, default_branch):
+    # websub/websubhub load tests are in websubhub module, hence `websub` load-test badge should be same as `websubhub` load-test badge
+    if module_name == "module-ballerina-websub":
+        module_name = "module-ballerina-websubhub"
+        default_branch = "main"
```
dilanSachi, line 101: why do we have main as the default branch in websubhub(not related to this PR though)

ayeshLK: When we initialize the repository from GitHub side it will create `main` branch and mark it as the default branch.

dilanSachi (approved):

release/src/dependencies/utils.py
ThisaruGuruge: Why do we have to provide the default branch separately here? The default branch is inherited from the GitHub APIs.

ayeshLK: Logic here is; 
- if the `module_name` is `module-ballerina-websub`, it should use the same `load-tests-badge` as `websubhub`

Hence we update the `module_name` to `module-ballerina-websubhub`. But, since the original module is `websub` the `default branch` which is inherited from the GitHub API is  `master`. But since we are updating the module to `websubhub` we should make sure the `default_branch` is updated to the default-branch of `websubhub` which is `main`.

dilanSachi: Apparently, even though the url has `main` or `master` it correctly points to the workflow file. But better to explicitly mention the default_branch.

ThisaruGuruge: You don't need the branch to access the workflow. It requires only if we need to get the workflow from a branch other than the default branch. In our case, it's always the default branch. So no need to provide the branch here.

dilanSachi: so you mean we dont need to add `main` in `https://raw.githubusercontent.com/ballerina-platform/module-ballerina-websubhub/main/.github/workflows/process-load-test-result.yml`?

ThisaruGuruge: Why do we need the raw file of the workflow here? We only need the workflow link right?

dilanSachi: We check whether the workflow file exists to identify whether the module contains load tests (to show the N/A badge for those that dont have load tests).

ThisaruGuruge: Oh, to identify whether a workflow exists right? Can't we check that using the GitHub API call? It should return an error when we try to get the badge?

My concern here mainly is, we are special-casing a single repo here, and to handle that, we have to change the function definition. Maybe we can introduce a local variable in that case. 

The modules.json anyways contain the default branch of each branch. Hard-coding the branch name is not the ideal way to fix this.

dilanSachi: I will try the GitHub API call suggestion

ThisaruGuruge: Maybe an easy approach would be to pass the `module` instead of the `module_name` to the `get_load_tests_badge()` function and retrieve the default branch inside the function, instead of hard-coding this.

This is working only because all the `master` branch requests will be redirected to `main` if the `master` branch does not exist in a repo. But it doesn't work for the raw file access.

Issue #3581, opened 2022-11-01
Use Workflow Templates for Release and Pull Request Builds


ThisaruGuruge:
**Description:**
$Subject.

To use the workflow templates, follow these examples:

The `publish-release.yml` workflow:

```yaml
name: Publish Release
on:
  workflow_dispatch:
  repository_dispatch:
    types: [ stdlib-release-pipeline ]
jobs:
  release-package:
    name: Release Package
    uses: ballerina-platform/ballerina-standard-library/.github/workflows/release-package-template.yml@main
    with:
      package-org: "ballerina" # Required
      package-name: "graphql" # Required
      additional-build-flags: "-x :graphql-examples:build" # Optional - Used for the step `Build without Tests`
      additional-publish-flags: "" # Optional - Used for the step `Publish Package`
```

The `pull-request.yml`:

```yaml
name: Pull Request Build
on: pull_request
jobs:
  pull-request-build:
    name: Pull Request Build
    uses: ballerina-platform/ballerina-standard-library/.github/workflows/pull-request-build-template.yml@main
    with:
      additional-ubuntu-test-flags: "-P disable=disabled_on_ubuntu" # Optional - Used for running/skipping a group of tests in Ubuntu
      additional-windows-test-flags: "-P disable=disabled_on_windows" # Optional - Used for running/skipping a group of tests in Windows
```

Module List:

- [ ] constraint
- [ ] io
- [ ] java.arrays
- [ ] regex
- [ ] time
- [ ] url
- [ ] xmldata
- [ ] crypto
- [ ] log
- [ ] os
- [ ] protobuf
- [ ] random
- [ ] serdes
- [ ] task
- [ ] xslt
- [ ] cache
- [ ] file
- [ ] ftp
- [ ] mime
- [ ] tcp
- [ ] udp
- [ ] uuid
- [ ] auth
- [ ] email
- [ ] jwt
- [ ] oauth2
- [ ] http
- [ ] grpc
- [ ] websocket
- [ ] websub
- [ ] websubhub
- [ ] graphql
- [ ] sql
- [ ] nats
- [ ] stan
- [ ] kafka
- [ ] rabbitmq
- [ ] java.jdbc
- [ ] mssql
- [ ] mysql
- [ ] oracledb
- [ ] postgresql

daneshk:
This is already done. Hence, closing the issue

PR #3584, opened 2022-11-01
Add common release workflow file to reuse

ThisaruGuruge:
## Purpose

This PR introduces two workflows that can be reused by the standard library repositories. The `Publish Release` workflow and the `Pull Request Build` workflow. Currently, these workflows reside in each of the standard library package repositories, by duplicating almost the same code. The drawback with this approach is that when there's a change is required, each and every repo has to do the change separately, which is cumbersome. 

In the long run, this will help to update the workflow easily, since we don't have to update the workflow in each and every repo. Instead, we can update the workflow in this repository.

This also includes a fix for concurrent builds, so that when a PR branch has a new commit, the already running PR builds will cancel and a new workflow run starts. This helps to reduce the workflow runs, which causes other workflows to queue due to heavy load.

Related: #3581 

The following is an example of reusing the release workflow:

```yaml
name: Publish Release
on:
  workflow_dispatch:
  repository_dispatch:
    types: [ stdlib-release-pipeline ]
jobs:
  release-package:
    name: Release Package
    runs-on: ubuntu-latest
    steps:
    - uses: ballerina-platform/ballerina-standard-library/.github/workflows/release-package-template.yml@main
      with:
        package-org: "ballerina" # Required
        package-name: "graphql" # Required
        additional-build-flags: "-x :graphql-examples:build" # Optional - Used for the step `Build without Tests`
        additional-publish-flags: "" # Optional - Used for the step `Publish Package`
```

The following is an example of using the pull request build workflow:

```yaml
name: Pull Request Build
on: pull_request
jobs:
  release-package:
    name: Release Package
    runs-on: ubuntu-latest
    steps:
    - uses: ballerina-platform/ballerina-standard-library/.github/workflows/pull-request-build-template.yml@main
      with:
        additional-ubuntu-test-flags: "-P disable=disabled_on_ubuntu" # Optional - Used for running/skipping a group of tests in Ubuntu
        additional-windows-test-flags: "-P disable=disabled_on_windows" # Optional - Used for running/skipping a group of tests in Windows
```


.github/workflows/release_package.yml
```diff
@@ -0,0 +1,112 @@
...
+    if: github.repository_owner == 'ballerina-platform'
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v2
+
+      - name: Set up JDK 11
+        uses: actions/setup-java@v2
+        with:
+          distribution: "temurin"
+          java-version: 11
```
niveathika, line 32: Wouldnt this checkout `ballerina-standard-library` instead of the required module?

.github/workflows/release_package.yml
```diff
@@ -0,0 +1,112 @@
...
+          GITHUB_TOKEN: ${{ secrets.BALLERINA_BOT_TOKEN }}
+          BALLERINA_CENTRAL_ACCESS_TOKEN: ${{ secrets.BALLERINA_CENTRAL_DEV_ACCESS_TOKEN }}
+          packageUser: ${{ secrets.BALLERINA_BOT_USERNAME }}
+          packagePAT: ${{ secrets.BALLERINA_BOT_TOKEN }}
+          publishUser: ${{ secrets.BALLERINA_BOT_USERNAME }}
+          publishPAT: ${{ secrets.BALLERINA_BOT_TOKEN }}
+        run: |
+          ./gradlew -Pversion=${VERSION} build -x test -PpublishToCentral=true ${{ inputs.additional-publish-flags }}
+
+      - name: Publish to Ballerina Central Stage
```
niveathika, line 96: I am -1 on having publishing to dev and stage in the release build. Dev and stage are not as resilient as prod. Publish tend to fail 40-60% of the time. This can lead to delays during release time. 

These steps are needed only for patch releases, releases concurrently to the distribution uses, https://github.com/ballerina-platform/ballerina-release/actions/workflows/sync_dev_and_staging_central_on_release.yml. This is much easier than us trying to sync every release

.github/workflows/release_package.yml
ThisaruGuruge: No, when the `actions/checkout` is used in a called workflow, it will be run as if it is invoked in the caller workflow.

Refer the `Overview` section here: https://docs.github.com/en/actions/using-workflows/reusing-workflows#overview

.github/workflows/release_package.yml
ThisaruGuruge: +1. Will remove this.

.github/workflows/release_package.yml
niveathika: Ack

niveathika (approved):

niveathika (approved):

PR #3903, opened 2023-01-05
Make `java_tool_options` to be optional in GraalVM Check

TharmiganK:
## Purpose

This env variable is set by default on windows runner since some of the standard libraries need this property to run the tests.

This PR will make setting `JAVA_TOOL_OPTIONS` env variable optional since having this as a default option for windows run is creating unexpected failures(See the issue below).

Fixes https://github.com/ballerina-platform/ballerina-standard-library/issues/3899


.github/workflows/build-with-bal-test-native-template.yml
```diff
@@ -26,6 +26,10 @@ on:
                 required: false
                 type: string
                 default: ''
+            java_tool_options:
+                required: false
+                type: string
+                default: ''
```
ThisaruGuruge, line 32: Shall we make the default to UTF-8?

ThisaruGuruge: My bad, we don't need that here, right?

ThisaruGuruge (approved):
LGTM

PR #4003, opened 2023-01-27
Improve the Standard Library Full Build Script

ThisaruGuruge:
## Purpose

Fixes: https://github.com/ballerina-platform/ballerina-standard-library/issues/4002

This will improve the Standard Library Build script with the following:

- There was a bug that prevented force-checkout the repositories. It is now fixed.
- When a full build is running, the SNAPSHOT build does not change the version number. It just replace the timestamp version with `SNAPSHOT`. This may cause issues when we build the full standard library using the script.
- Introduce `--from-module` flag: This will help to start the build from a particular module.
- Change the `--module` flag to `--up-to-module` to make the purpose of the flag clear.
- Change the warning color to Yellow from Red.
- Introduce `--continue-on-error` flag and make the default behavior to stop the subsequent builds when a module build is failed. The existing behavior is to continue on error, no matter what.

### DIFF
--- a/resources/scripts/build_standard_library.py
+++ b/resources/scripts/build_standard_library.py
@@ -24,26 +24,44 @@ TEMP_PROPERTIES = "temp.properties"
 GRADLE_PROPERTIES = "gradle.properties"
 
 # Argument Parser
-parser = argparse.ArgumentParser(description='Incrementally Build the Ballerina Standard Library')
+parser = argparse.ArgumentParser(
+    description='Incrementally Build the Ballerina Standard Library')
 
 # Mandatory Arguments
-parser.add_argument('path', help='Path to the directory where the standard library modules are (need to be) cloned')
+parser.add_argument(
+    'path', help='Path to the directory where the standard library modules are (need to be) cloned')
 
 # Optional arguments
-parser.add_argument('--lang-version', help='Ballerina language version to use for the builds')
-parser.add_argument('--branch', help='Branch to build. (Build will fail if the branch not found')
-parser.add_argument('--snapshots-build', action="store_true", help="Replace all the standard library dependent versions with 'SNAPSHOT' versions. This is helpfull to incrementally build libraries on top of a local change")
-parser.add_argument('--publish-to-local-central', action="store_true", help="Publish all the modules to the local ballerina central repository")
-parser.add_argument('--skip-tests', action="store_true", help="Skip tests in the builds")
-parser.add_argument('--keep-local-changes', action="store_true", help="Stop updating the repos from the origin. Keep the local changes")
-parser.add_argument('--module', help="Build up to the specified module")
-parser.add_argument('--module-list', help="Path to the module list JSON file. If not provided, the existing 'stdlib_modules.json' from the GitHub will be used")
-parser.add_argument('--build-distribution', action="store_true", help="If the distribution should be built on top of the changes been done. This has to be used with '--snapshots-build' to be affective")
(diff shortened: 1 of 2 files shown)


resources/scripts/build_standard_library.py
```diff
@@ -208,17 +282,31 @@ def clone_module(module_link):
...
+        for line in properties:
+            if re.match("^version=", line):
+                module_name = module[FIELD_NAME].split("-")[-1]
+                if module_name == "jballerina.java.arrays":
+                    module_name = "JavaArrays"
+                elif module_name == "oauth2":
+                    module_name = "OAuth2"
+                else:
+                    module_name = module_name.capitalize()
+                version_dict[module_name] = line.split("=")[1].strip()
```
chiranSachintha, line 302: We can use `version_key` to update the dictionary. (The version key of the constant module is  `stdlibConstraintVersion`)
```python
version_dict[module['version_key']] = module['module_version']

```

resources/scripts/build_standard_library.py
```diff
@@ -187,13 +250,24 @@ def replace_lang_version(version):
...
 
-def replace_stdlibs_version():
+def replace_stdlibs_version(module_name, snapshots_build):
+    global version_dict
     with open(TEMP_PROPERTIES, "w+") as temp, open(GRADLE_PROPERTIES) as properties:
         for line in properties:
-            if re.match("^stdlib.*Version=", line):
-                version_string = line.split("=")[0]
-                version_number = line.split("=")[1].split("-")[0]
-                line = version_string + "=" + version_number + "-SNAPSHOT" + "\n"
```
chiranSachintha, line 196: By keeping it as 'stdlibConstraintVersion', we can directly access the version number from the dictionary using the 'version_string'.
```
line = version_string + "=" + version_dict[version_string]  + "\n"
```

resources/scripts/build_standard_library.py
```diff
@@ -208,17 +282,31 @@ def clone_module(module_link):
...
+                module_name = module[FIELD_NAME].split("-")[-1]
+                if module_name == "jballerina.java.arrays":
+                    module_name = "JavaArrays"
+                elif module_name == "oauth2":
+                    module_name = "OAuth2"
+                else:
+                    module_name = module_name.capitalize()
+                version_dict[module_name] = line.split("=")[1].strip()
+
+
```
chiranSachintha, line 304: This extra line can be removed

resources/scripts/build_standard_library.py
```diff
@@ -156,26 +216,29 @@ def process_module(module, commands, lang_version, use_snapshots):
...
-        replace_stdlibs_version()
+        replace_stdlibs_version(module[FIELD_NAME], use_snapshots)
 
-    subprocess.run(commands)
+    proc = subprocess.run(commands)
 
     os.chdir("..")
 
+    return proc.returncode
+
```
chiranSachintha, line 241:
```suggestion
```

resources/scripts/build_standard_library.py
ThisaruGuruge: These extra lines are added by the Python formatter. (The convention is to use two empty lines between two function definitions).

resources/scripts/build_standard_library.py
ThisaruGuruge: Fixed with 33d788f226ccf744b0fe702ece2da53c1259f4f7

resources/scripts/build_standard_library.py
ThisaruGuruge: Fixed with 33d788f226ccf744b0fe702ece2da53c1259f4f7

niveathika (approved):

PR #6284, opened 2024-04-04
Add data.jsondata and data.xmldata packages to Ballerina Library pipeline

prakanth97:
## Purpose
$title


ThisaruGuruge (commented):
Let's define version keys for these modules since they have special characters in the module name.

prakanth97:
> Let's define version keys for these modules since they have special characters in the module name.

Updated

ThisaruGuruge (approved):

PR #6869, opened 2024-08-14
Fix process load test results workflow

TharmiganK:
## Purpose

The load-tests has been failing for some time due to the following issue:
- The merge PR step hangs intermittently and get timed-out, which makes the PR not merged until the next load test results PR
- Due to the above reason, there is already a `load-test-results` branch which is not sync with the master, which makes the push branch step fail

In this PR, the above issues has been addressed with the following changes:
- Use a github CLI command to wait until all the required checks pass
- Check for already existing branch and remove it before pushing the new changes - Ideally this check is not required when the PR is merged automatically

### DIFF
--- a/.github/workflows/process-load-test-results-template.yml
+++ b/.github/workflows/process-load-test-results-template.yml
@@ -58,9 +58,21 @@ jobs:
           done
         env:
           GITHUB_TOKEN: ${{ secrets.ballerina_bot_token }}
+      - name: Check if the load-test-results branch exists
+        if: ${{ steps.processResults.outputs.hasPassed }}
+        id: check_branch
+        run: |
+          if git ls-remote --heads origin load-test-results | grep -q "refs/heads/load-test-results"; then
+            echo "branch_exists=true" >> $GITHUB_ENV
+          else
+            echo "branch_exists=false" >> $GITHUB_ENV
+          fi
+      - name: Delete existing load-test-results branch
+        if: ${{ steps.check_branch.outputs.branch_exists }}
+        run: git push origin --delete load-test-results
       - name: Push results
         if: ${{ steps.processResults.outputs.hasPassed }}
-        run: git push origin "load-test-results"
+        run: git push origin load-test-results
         env:
           GITHUB_TOKEN: ${{ secrets.ballerina_bot_token }}
       - name: Create pull request for new summary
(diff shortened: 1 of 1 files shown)


.github/workflows/process-load-test-results-template.yml
```diff
@@ -58,9 +58,21 @@ jobs:
           done
         env:
           GITHUB_TOKEN: ${{ secrets.ballerina_bot_token }}
+      - name: Check if the load-test-results branch exists
```
TharmiganK, line 61:
```suggestion
      - name: Check the existence of load-test-results branch
```

.github/workflows/process-load-test-results-template.yml
```diff
@@ -58,9 +58,21 @@ jobs:
...
+        if: ${{ steps.processResults.outputs.hasPassed }}
+        id: check_branch
+        run: |
+          if git ls-remote --heads origin load-test-results | grep -q "refs/heads/load-test-results"; then
+            echo "branch_exists=true" >> $GITHUB_ENV
+          else
+            echo "branch_exists=false" >> $GITHUB_ENV
+          fi
+      - name: Delete existing load-test-results branch
```
TharmiganK, line 70:
```suggestion
      - name: Delete the existing load-test-results branch
```

.github/workflows/process-load-test-results-template.yml
```diff
@@ -58,9 +58,21 @@ jobs:
           done
         env:
           GITHUB_TOKEN: ${{ secrets.ballerina_bot_token }}
+      - name: Check the existence of load-test-results branch
+        if: ${{ steps.processResults.outputs.hasPassed }}
```
NipunaRanasinghe, line 62: Can there be any cases where the `load-test-results` branch is created even if the `hasPassed` is set to false? (If the previous step fails inbetween those 2 steps it can happen IMO). So is that an option to clear the existing branch regardless of the output from the previous step?

TharmiganK: In the previous step we create a branch locally but we push the branch in the next step. We only push the changes when the load tests are passing, otherwise we will not push. So I have added this branch removal with the same check, so that it will be removed only if we want to push the same branch in the next step

NipunaRanasinghe: got it. makes sense

NipunaRanasinghe (commented):
LGTM

ThisaruGuruge:
> Check for already existing branch and remove it before pushing the new changes - Ideally this check is not required when the PR is merged automatically

I am -1 for this. We have discussed this previously as well. If a load test PR was failed due to a GitHub issue, we will lose that load test result because of this. The recommended way is to module owners to manually handle this (or maybe we can come up with a way to automate this, but I'm not sure). 

Manually handling:
1. Close the failed PR
2. Delete the branch
3. Re-run the failed load test processing workflows

TharmiganK:
> I am -1 for this. We have discussed this previously as well. If a load test PR was failed due to a GitHub issue, we will lose that load test result because of this. The recommended way is to module owners to manually handle this (or maybe we can come up with a way to automate this, but I'm not sure).

Ack, will remove the deleting part.

May be we can send a notification to a notification chat group with the  library team members, so that they can address it asap

.github/workflows/process-load-test-results-template.yml
```diff
@@ -60,7 +60,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.ballerina_bot_token }}
       - name: Push results
         if: ${{ steps.processResults.outputs.hasPassed }}
-        run: git push origin "load-test-results"
+        run: git push origin load-test-results
```
TharmiganK, line 63:
```suggestion
        run: git push origin "load-test-results"
```

ThisaruGuruge (approved):
LGTM

TharmiganK:
Since the workflows are failing currently in the merge PR step, I will merge this change and test. Sending notification on failure will be addressed separately

PR #7276, opened 2024-10-18
Update connector workflow templates to use ballerina-bot token as the `packagePAT`

ayeshLK:
This change is required to build the Ballerina IBM CTG connector [1] as some of the dependent resources has been published under the `wso2-enterprise` organization and `github.token` does not have access to read packages from a private organization.

[1] - https://github.com/ballerina-platform/module-ballerinax-ibm.ctg

### DIFF
--- a/.github/workflows/build-connector-template.yml
+++ b/.github/workflows/build-connector-template.yml
@@ -67,8 +67,8 @@ jobs:
 
             -   name: Build the Package
                 env:
-                    packageUser: ${{ github.actor }}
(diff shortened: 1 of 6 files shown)


ayeshLK:
The `packagePAT` should come in as a parameter to the workflow-template and it should be overridden by only the workflows which really requires it [1] Hence, closing this PR.

[1] - https://github.com/ballerina-platform/ballerina-library/issues/7284

PR #7561, opened 2025-02-07
[Main] Update CodeCov Report Publication 

ThisaruGuruge:
## Purpose
$Subject


Nuvindu (approved):

.github/workflows/release_pipeline.yml
```diff
@@ -18,6 +18,11 @@ on:
                 required: true
                 default: false
                 type: boolean
+            release_driver_modules:
+                description: 'Release Ballerina Driver Modules (ballerinax/mysql.driver, ballerinax/mssql.driver, etc.)'
+                required: true
+                default: false
+                type: boolean
```
daneshk, line 25: Driver modules are not released in every release. It is mainly on-demand releases.

daneshk (approved):
Add a minor comment. LGTM

.github/workflows/release_pipeline.yml
ThisaruGuruge: This is for cases where we have to release all in a single go, and I agree, that isn't the main case. But for completeness, I'd be happy to keep this option.

daneshk (approved):

Issue #7803, opened 2025-04-17
Connector Regeneration Workflow Should Only Check the PR Build Status


ThisaruGuruge:
### Current Limitation

Currently, when a connector regeneration workflow runs, the auto-merge functionality depends on all checks being passed. This is not necessary. Reasons are below:

* CLA Assistant: We do not need CLA check since the Ballerina bot and the reviewer bot already have signed it.
* Code Coverage Checks: We are not checking the code coverage for generated connectors

### Suggested Improvement

The auto-merge feature should only check the PR build status, and if the build is passing, we should merge it.

### Version

N/A

ThisaruGuruge:
@chathushkaayash FYI

PR #7805, opened 2025-04-17
Update auto PR merge step to only check for PR Build Status in Regenerate OpenAPI Connectors Workflow

chathushkaayash:
Resolves [Connector Regeneration Workflow Should Only Check the PR Build Status ](https://github.com/ballerina-platform/ballerina-library/issues/7803)

### DIFF
--- a/.github/workflows/regenerate-connector-template.yml
+++ b/.github/workflows/regenerate-connector-template.yml
@@ -236,7 +236,6 @@ jobs:
 
           while [[ $retry_count -lt $max_retries ]]
           do
-            sleep 20
             retry_count=$((retry_count + 1))
             
             echo "[$retry_count/$max_retries] Checking PR Build Workflow status..."
@@ -246,6 +245,7 @@ jobs:
               --json statusCheckRollup)
 
             if [[ -z "$result" ]]; then
+              sleep 20
               continue
             fi
 
--- a/.github/workflows/regenerate-connector-template.yml
+++ b/.github/workflows/regenerate-connector-template.yml
@@ -231,15 +231,24 @@ jobs:
       - name: Merge PR
         if: ${{ steps.commitFiles.outputs.hasChanged == 'true' && inputs.auto-merge == true }}
         run: |
-          checkCount="0"
-          while [ "$checkCount" != "1" ]
+          while true
           do
             sleep 20
-            checkCount=$(gh pr view ${{ steps.createPR.outputs.prUrl }} --jq '[.statusCheckRollup[] | select((.conclusion=="SUCCESS") and ((.name=="Run PR Build Workflow / Build")))] | length' --json statusCheckRollup)
-            failedCount=$(gh pr view ${{ steps.createPR.outputs.prUrl }} --jq '[.statusCheckRollup[] | select((.conclusion=="FAILURE") and ((.name=="Run PR Build Workflow / Build")))] | length' --json statusCheckRollup)
-            if [[ "$failedCount" != "0" ]]
-            then
-              echo "PR Build has Failed"
+            result=$(gh pr view ${{ steps.createPR.outputs.prUrl }} \
+              --jq '[.statusCheckRollup[] | select((.name == "Run PR Build Workflow / Build") and (.status == "COMPLETED"))][0]' \
+              --json statusCheckRollup)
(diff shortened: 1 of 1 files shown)


.github/workflows/regenerate-connector-template.yml
```diff
@@ -231,25 +231,17 @@ jobs:
...
-
-            if [[ "$failedCount" -gt 0 ]]; then
-              echo "Some status checks failed or were cancelled."
+            checkCount=$(gh pr view ${{ steps.createPR.outputs.prUrl }} --jq '[.statusCheckRollup[] | select((.conclusion=="SUCCESS") and ((.name=="Run PR Build Workflow / Build")))] | length' --json statusCheckRollup)
+            failedCount=$(gh pr view ${{ steps.createPR.outputs.prUrl }} --jq '[.statusCheckRollup[] | select((.conclusion=="FAILURE") and ((.name=="Run PR Build Workflow / Build")))] | length' --json statusCheckRollup)
+            if [[ "$failedCount" != "0" ]]
+            then
+              echo "PR Build has Failed"
               exit 1
             fi
```
ThisaruGuruge, line 244: I think we can remove the `checkCount` as we don't need any count here. Instead, we can just get the status of the workflow;
* If it's `SUCCESS`, we break the loop and mark this as passed.
* If it's `FAILURE`,`CANCELLED`, `SKIPPED`, `STALE`, or `TIMED_OUT`, we mark this as failed.

We can also check the `status` also before the conclusion to see if the `status` is completed. (This isn't necessary, but looks clean though).

chathushkaayash: Sure, updated the workflow with [b692491940788e3c772af1a7267867197296e78c](https://github.com/ballerina-platform/ballerina-library/pull/7805/commits/b692491940788e3c772af1a7267867197296e78c)

.github/workflows/regenerate-connector-template.yml
```diff
@@ -231,24 +231,25 @@ jobs:
       - name: Merge PR
         if: ${{ steps.commitFiles.outputs.hasChanged == 'true' && inputs.auto-merge == true }}
         run: |
-          checkComplete="0"
-          while [ "$checkComplete" != "1" ]; do
+          while true
+          do
             sleep 20
```
ThisaruGuruge, line 236: Why do we need a sleep here? There are two sleep statements, and I think this one is redundant.

.github/workflows/regenerate-connector-template.yml
```diff
@@ -231,24 +231,25 @@ jobs:
       - name: Merge PR
         if: ${{ steps.commitFiles.outputs.hasChanged == 'true' && inputs.auto-merge == true }}
         run: |
-          checkComplete="0"
-          while [ "$checkComplete" != "1" ]; do
+          while true
+          do
```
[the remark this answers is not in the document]
chathushkaayash: Added a maximum  retry count with [35108ccf054a9dfa63913d7f61d07e8a568e471d](https://github.com/ballerina-platform/ballerina-library/pull/7805/commits/35108ccf054a9dfa63913d7f61d07e8a568e471d)

.github/workflows/regenerate-connector-template.yml
chathushkaayash: The first sleep in the while loop works as a polling interval between status checks, and the second sleep works as a short delay before proceeding to the next step once the loop has completed.

ThisaruGuruge: Usually, the sleep should be at the end of the loop, instead of the beginning of the loop. A minor detail, but I think we should follow that. 

The reason behind this is that we should sleep after a failure, not before checking something failed or not. (In case of a success scenario, we should be able to attend to that immediately, instead of waiting redundantly.

chathushkaayash: Update to sleep only when the result is not complete [4ac135a54cfc2b906bc7c8cf150cf7008f7364ab](https://github.com/ballerina-platform/ballerina-library/pull/7805/commits/4ac135a54cfc2b906bc7c8cf150cf7008f7364ab)

ThisaruGuruge (approved):

PR #8440, opened 2025-11-06
Add Auto generate Connector Template

nostoc:

### DIFF
--- a/.github/workflows/auto_generate_connector_template.yml
+++ b/.github/workflows/auto_generate_connector_template.yml
@@ -12,16 +12,6 @@ on:
         required: false
         type: string
         default: "nightly"
-      use-pipeline:
-        description: "Runs all steps of the connector generation process"
-        required: false
-        type: boolean
-        default: true
-      generation-path:
-        description: "Path where Ballerina client should be generated"
-        required: false
-        type: string
-        default: "ballerina"
       license-file:
         description: "Path to license file relative to docs directory"
         required: false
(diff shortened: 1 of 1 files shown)


.github/workflows/auto_generate_connector_template.yml
```diff
@@ -0,0 +1,434 @@
...
+        description: "Path where Ballerina client should be generated"
+        required: false
+        type: string
+        default: "ballerina"
+      license-file:
+        description: "Path to license file relative to docs directory"
+        required: false
+        type: string
+        default: "license.txt"
+      auto-confirm:
```
TharmiganK: In workflows, can we have this as false?

.github/workflows/auto_generate_connector_template.yml
```diff
@@ -0,0 +1,434 @@
+name: Auto-Generate Connector Template
+
+on:
+  workflow_call:
+    inputs:
+      openapi-url:
+        description: "URL of the OpenAPI specification (JSON/YAML)"
+        required: true
```
TharmiganK: This is not required if we have already committed the spec right?

.github/workflows/auto_generate_connector_template.yml
```diff
@@ -0,0 +1,216 @@
...
+      openapi-url:
+        description: "URL of the OpenAPI specification (JSON/YAML)"
+        required: false
+        type: string
+      ballerina-version:
+        description: "Ballerina version to use"
+        required: false
+        type: string
+        default: "nightly"
+      use-pipeline:
```
TharmiganK: We don't need this anymore right?

Issue #8510, opened 2025-12-08
AI Driven Dependabot for OpenAPI Connectors


TharaniDJ:
### Area

Area/Connector

### Description

Currently, Ballerina supports more than 500 connectors, from which more than 300 are generated. Maintaining and keeping up-to-date such a large number of connectors is a cumbersome task and needs a lot of manual intervention, which is not practical. Therefore, we need a “Dependabot-like” bot that checks for newer APIs and updates the relevant connector repositories. But keeping track of OpenAPI specifications is not straightforward as maven dependencies. Therefore, we need an AI-driven solution for this purpose.

TharaniDJ:
The project proposal is completed and currently in the internal review stage.

TharaniDJ:
The OpenAPI spec update detection is complete for GitHub- published and verified specs. The connector generation and version analysis are also mostly done.

TharaniDJ:
## **Progress update: [Phase 01]**
I have completed the full pipeline implementation for Phase 1. As discussed, Phase 1 focuses on GitHub-published and verified OpenAPI specifications. The system now supports the following end-to-end workflow:

**OpenAPI Dependabot Workflow:**

The workflow runs on a scheduled basis using GitHub Actions (cron: 0 2 * * *, daily at 2 AM UTC) and can also be triggered manually via workflow_dispatch. This ensures both automated monitoring and on-demand execution when needed.

1. Detects updates to the relevant OpenAPI specifications
The update detection is implemented in Ballerina using the ballerinax/github client, supporting both release-tag–based detection (via /releases/latest, tag comparison, and SHA-256 hashing) and file-based detection (branch spec download with info.version extraction from the OpenAPI document).

2. Downloads the updated openapi specifications and creates pull requests to the https://github.com/wso2/api-specs repository 

3. After merging the pull request to the https://github.com/wso2/api-specs repository, trigger the connector auto-generation workflow in the corresponding connector repositories

4. Regenerates connectors using the existing pipeline (with minor enhancements to properly support regeneration scenarios)

5. Performs version analysis for the generated Ballerina connector and updates the relevant version files (Gradle properties file, etc.)
[...]

PR #8765, opened 2026-04-27
Automate connector regeneration

TharaniDJ:
After identifying latest openapi specs from openapi dependabot workflow this will regenerate existing connectors with latest specs

## Purpose
Issue - https://github.com/ballerina-platform/ballerina-library/issues/8510

## Goals
Sets a flag to identify that this is a connector regeneration in the connector automator workflow
A new workflow to do the version analysis of the connector 

## Summary

This pull request automates regenerating Ballerina connectors from updated OpenAPI specifications and extends the connector automator with controlled regeneration, recovery, version-analysis, and orchestration tooling to streamline connector updates and reporting.

### Key changes

- Orchestration workflow
  - Adds a GitHub Actions workflow (.github/workflows/orchestrate-connector-updates.yml) that accepts connector update payloads (repository_dispatch or manual input), persists connector lists, dispatches per-repo update events, monitors resulting “Auto-generated connector” PRs, aggregates PR metadata (including inferred semantic-change classification), uploads artifacts, and sends a consolidated HTML email summary.

- Regeneration flow and recovery
  - Adds a `regenerate` pipeline mode to the automator CLI that reapplies saved sanitations, re-sanitizes specs, regenerates code/tests/docs, and implements a multi-phase recovery process when builds fail (LLM-driven code fixes, test/example regeneration, retry logic).
  - Introduces a flag to mark workflow runs as connector regenerations so orchestration can distinguish automated regenerations.

- Version analysis and AI integration
  - Adds a version-analysis module that classifies required semantic version bumps (MAJOR/MINOR/PATCH) from diffs using Anthropic’s API and persists analysis results.
  - Updates AI provider selection and adds a dependency on ballerinax:ai.anthropic v1.3.0; updates dependency lock entries and removes a pinned distribution identifier to allow flexible resolution.

- Developer tooling and test-generation improvements
  - Adds utilities to sort Ballerina client methods and type declarations, update changelogs from PR descriptions, and enrich test generation with full client and type contexts.
  - Enhances the code-fixer to maintain per-file fix history and include type/context information for iterative LLM-driven fixes.
  - Extends the example generator to attempt recovery of existing examples before regenerating them.

- Sanitization management
  - Introduces a sanitations handler to generate and reapply sanitation rules (server URLs, path prefixes, format adjustments, nullability/type updates) and persist sanitations.md for reuse across regenerations.

- Documentation and other edits
  - Removes connector_automator/README.md and connector_automator/docs/spec.md (documentation appears reorganized or relocated).
  - Adds multiple new modules and helper utilities and makes extensive implementation updates within the connector_automator module.

### Dependency and configuration updates

- Adds ballerinax:ai.anthropic v1.3.0 and updates connector_automator/Dependencies.toml to newer package/distribution versions.
- Removes a previously pinned distribution identifier from connector_automator/Ballerina.toml.

### Outcomes

- Enables end-to-end automation for regenerating connectors when OpenAPI specs are updated, including orchestration, monitoring, recovery, version analysis, and consolidated reporting to help maintainers review and act on generated connector changes.


.github/workflows/orchestrate-connector-updates.yml
```diff
@@ -0,0 +1,468 @@
...
+on:
+  repository_dispatch:
+    types: [orchestrate-connector-updates]
+  workflow_dispatch:
+    inputs:
+      connectors_json:
+        description: 'JSON array of connector updates (for manual testing)'
+        required: false
+        type: string
+        default: '[]'
```
ThisaruGuruge: This part is not clear. What should be the input?

.github/workflows/orchestrate-connector-updates.yml
```diff
@@ -0,0 +1,468 @@
...
+            echo "Triggering: $REPO"
+            echo "  Spec: $SPEC"
+            echo "  Version: $VERSION"
+            echo "  URL: $OPENAPI_URL"
+
+            if gh api \
+              repos/"$REPO"/dispatches \
+              -X POST \
+              -f event_type=openapi-update \
+              -f client_payload[sender]=orgBBalLib/ballerina-library \
```
ThisaruGuruge: Check all the references to the `orgBBalLib`

```suggestion
              -f client_payload[sender]=ballerina-platform/ballerina-library \
```

.github/workflows/orchestrate-connector-updates.yml
```diff
@@ -0,0 +1,468 @@
...
+          GH_TOKEN: ${{ secrets.BALLERINA_BOT_TOKEN }}
+        run: |
+          echo "Waiting for connector updates to process..."
+          echo "Looking for PRs created after: ${{ steps.start_time.outputs.workflow_start_time }}"
+
+          TOTAL_CONNECTORS=$(wc -l < triggered_connectors.txt)
+          echo "Monitoring $TOTAL_CONNECTORS connector repositories..."
+
+          INITIAL_WAIT=600
+          echo "⏳ Waiting ${INITIAL_WAIT} seconds (10 minutes) for connector generation workflows to process..."
```
ThisaruGuruge: Let's remove the emojis from the logs (and other places).

.github/workflows/orchestrate-connector-updates.yml
```diff
@@ -0,0 +1,468 @@
...
+        id: send_email
+        uses: dawidd6/action-send-mail@v3
+        with:
+          server_address: smtp.gmail.com
+          server_port: 465
+          secure: true
+          username: ${{ secrets.GMAIL_USERNAME }}
+          password: ${{ secrets.GMAIL_APP_PASSWORD }}
+          subject: ${{ steps.generate_email.outputs.email_subject }}
+          to: t37714687@gmail.com
```
ThisaruGuruge: `to` should have a list of emails. They should be retrieved through secrets.

.github/workflows/orchestrate-connector-updates.yml
```diff
@@ -0,0 +1,468 @@
...
+          from: ${{ secrets.GMAIL_USERNAME }}
+          html_body: file://email_body.html
+
+      - name: Log email status
+        if: steps.trigger_connectors.outputs.triggered_count > 0
+        run: |
+          if [ "${{ steps.send_email.outcome }}" == "success" ]; then
+            echo "✅ Email sent successfully to t37714687@gmail.com"
+          else
+            echo "⚠️ WARNING: Email sending failed"
```
ThisaruGuruge: In each of these steps, let's maintain an env. variable to keep the workflow state. If any of the step failed, we should mark the workflow as failed (exit with non-zero).

P.S.: Workflow should mark as failed if it failed to trigger workflows, failed to send emails, etc. (Individual connector generation shouldn't be considered as a failure).

connector_automator/docs/spec.md
ThisaruGuruge, line 1: Why are we removing this file?

connector_automator/modules/client_regenerator/analyze_version_change.bal
```diff
@@ -0,0 +1,183 @@
...
+};
+
+function analyzeWithAnthropic(string gitDiff) returns AnalysisResult|error {
+    string apiKey = os:getEnv("ANTHROPIC_API_KEY");
+
+    if apiKey == "" {
+        return error("ANTHROPIC_API_KEY environment variable is not set");
+    }
+
+    io:println(string `Using Anthropic API (${apiKey.length()} chars)`);
```
ThisaruGuruge: We don't need this log. Check similar places.

connector_automator/modules/client_regenerator/analyze_version_change.bal
```diff
@@ -0,0 +1,183 @@
...
+
+    string prompt = string `You are analyzing git diff output for a Ballerina connector to determine the semantic version change needed.
+
+GIT DIFF:
+${gitDiff}
+
+RULES FOR VERSION CLASSIFICATION:
+- MAJOR: Breaking changes (removed/renamed methods, removed/renamed types, changed method signatures, changed field types, removed fields)
+- MINOR: Backward-compatible additions (new methods, new types, new optional fields)
```
ThisaruGuruge:
```suggestion
- MINOR: Backward-compatible additions (new methods, new types, new optional fields, new fields with default values)
```

connector_automator/modules/client_regenerator/analyze_version_change.bal
```diff
@@ -0,0 +1,183 @@
...
+- PATCH: Documentation changes, internal refactoring, bug fixes with no API surface changes
+
+Analyze the diff and respond with ONLY a JSON object (no markdown, no explanation):
+{
+  "changeType": "MAJOR|MINOR|PATCH",
+  "breakingChanges": ["list specific breaking changes"],
+  "newFeatures": ["list new features/additions"],
+  "bugFixes": ["list bug fixes or improvements"],
+  "summary": "concise summary of changes",
+  "confidence": 0.95
```
ThisaruGuruge: Should this be a number like `0.95` or something explains what `confidence` is?

connector_automator/modules/client_regenerator/analyze_version_change.bal
```diff
@@ -0,0 +1,183 @@
...
+}`;
+
+    http:Client anthropicClient = check new ("https://api.anthropic.com", {
```
ThisaruGuruge: Why can't we use existing [Anthropic connector](https://central.ballerina.io/ballerinax/ai.anthropic/1.3.2)?

BTW, the API URL is defined as a constant, but it was never used. We have a hard-coded URL here instead.

connector_automator/modules/client_regenerator/analyze_version_change.bal
```diff
@@ -0,0 +1,183 @@
...
+        httpVersion: http:HTTP_1_1,
```
ThisaruGuruge: Any specific reason to use HTTP 1.1?

connector_automator/modules/client_regenerator/analyze_version_change.bal
```diff
@@ -0,0 +1,183 @@
...
+        "max_tokens": 1024,
+        "temperature": 0.1,
+        "messages": [{"role": "user", "content": prompt}]
+    };
+
+    http:Request req = new;
+    req.setJsonPayload(payload);
+    req.setHeader("anthropic-version", "2023-06-01");
+    req.setHeader("x-api-key", apiKey);
+    req.setHeader("content-type", "application/json");
```
ThisaruGuruge: Won't need these if the connector is used.

connector_automator/modules/client_regenerator/analyze_version_change.bal
```diff
@@ -0,0 +1,183 @@
...
+
+    json response = {};
+    int retryCount = 0;
+    boolean success = false;
+
+    while !success && retryCount < MAX_RETRIES {
+        do {
+            http:Response httpResponse = check anthropicClient->post("/v1/messages", req);
+            int statusCode = httpResponse.statusCode;
+            io:println(string `Response status: ${statusCode}`);
```
ThisaruGuruge: Use typed responses instead of using `http:Response` object.

Refer: https://ballerina.io/learn/by-example/http-client-send-request-receive-response/

connector_automator/modules/client_regenerator/analyze_version_change.bal
```diff
@@ -0,0 +1,183 @@
...
+                if statusCode == 429 {
+                    retryCount = retryCount + 1;
+                    if retryCount < MAX_RETRIES {
+                        io:println(string `Rate limited. Retry ${retryCount}/${MAX_RETRIES} in ${RETRY_DELAY_SECONDS}s...`);
+                        runtime:sleep(RETRY_DELAY_SECONDS);
+                        continue;
+                    }
+                }
+                return error(string `Anthropic API returned status ${statusCode}: ${textResult}`);
+            }
```
ThisaruGuruge: You can configure [retry functionality on the client itself](https://ballerina.io/learn/by-example/http-retry/). No need to manually handle that.

connector_automator/modules/client_regenerator/analyze_version_change.bal
```diff
@@ -0,0 +1,183 @@
...
+
+    if gitDiffContent.length() == 0 {
+        return error("Git diff content is empty");
+    }
+
+    io:println("Analyzing with Claude...");
+
+    AnalysisResult analysis = check analyzeWithAnthropic(gitDiffContent);
+
+    io:println("\n" + repeatString("=", 60));
```
ThisaruGuruge: Can't we just define a constant string with `=====....` and use it? Are there any other usages of `repeatString` function?

connector_automator/modules/client_regenerator/sort_ballerina_client.bal
```diff
@@ -0,0 +1,253 @@
...
+    ContentBlock[] blocks = [];
+
+    int firstMethodLine = -1;
+    int lastMethodLine = -1;
+
+    int i = 0;
+    while i < lines.length() {
+        string line = lines[i];
+
+        if regex:matches(line, "\\s*resource\\s+isolated\\s+function\\s+(get|post|put|delete|patch)") {
```
ThisaruGuruge: This would match `resource isolated function` but not `isolated resource function` which is also correct syntax.

@lnash94 what's the signature we generate from the tool? Is it `resource isolated function`?

In any case, I think we should match both.

connector_automator/modules/client_regenerator/sort_ballerina_client.bal
```diff
@@ -0,0 +1,253 @@
...
+
+// Shared across client and type sort files within this module
+function countChar(string str, string char) returns int {
+    int count = 0;
+    foreach int i in 0 ..< str.length() {
+        if str.substring(i, i + 1) == char {
+            count += 1;
+        }
+    }
+    return count;
```
ThisaruGuruge: Can be replaced with:


```suggestion
    return re`${char}`.findAll(str).length()
```

.github/workflows/orchestrate-connector-updates.yml
```diff
@@ -0,0 +1,468 @@
+name: Orchestrate Connector Updates
```
lnash94, line 15: Shall we add a comment with a small description about the workflow?

connector_automator/modules/client_regenerator/sort_ballerina_client.bal
lnash94: For the Ballerina OpenAPI tool, we generate `resource isolated functions` and `remote isolated functions`. Remote functions are generated in scenarios where the endpoint path contains path segment parameters that cannot be handled using resource functions.

In such cases, we also need to match **remote** functions.

connector_automator/modules/client_regenerator/analyze_version_change.bal
```diff
@@ -0,0 +1,183 @@
...
+                runtime:sleep(RETRY_DELAY_SECONDS);
+            } else {
+                return e;
+            }
+        }
+    }
+
+    json contentJson = check response.content;
+    json[] content = check contentJson.ensureType();
+    string text = check content[0].text;
```
lnash94: Is there any possibility of retrieving more items in the `content`? Also, why does it always pick the first one?

.github/workflows/orchestrate-connector-updates.yml
TharaniDJ: This is a workflow_dispatch input for manual testing only the normal trigger is repository_dispatch. The input accepts a JSON array where each object has repository (the connector repo), specification (spec name), version (API version), and openapi_url (download URL for the spec). Updated the description with the full schema and an example.

.github/workflows/orchestrate-connector-updates.yml
TharaniDJ: Added WORKFLOW_STATUS env var initialized to success. Critical steps (trigger and email) set it to failed via $GITHUB_ENV if they fail. A final Fail workflow if critical step failed step checks the value and exits non-zero. Individual connector generation failures are recorded in the PR summary but don't affect WORKFLOW_STATUS

connector_automator/modules/client_regenerator/analyze_version_change.bal
TharaniDJ: The 0.95 was mistakenly hardcoded as a literal in the prompt template, so the LLM was copying it verbatim instead of reasoning about its own confidence. Changed confidence to a string field ("HIGH"/"MEDIUM"/"LOW") with a description in the prompt so the LLM actually assesses the clarity of the diff and picks the appropriate label

connector_automator/modules/client_regenerator/analyze_version_change.bal
TharaniDJ: No specific reason it was unnecessary. Removed along with the raw HTTP client since we now use the Anthropic connector via utils:callAI

connector_automator/modules/client_regenerator/analyze_version_change.bal
TharaniDJ: It was only used in this file. Replaced all three call sites with a const string SEPARATOR and removed the function.

connector_automator/modules/client_regenerator/analyze_version_change.bal
TharaniDJ: This code has been removed as part of addressing the other review comments. We've switched from the raw http:Client (where response.content was a JSON array and we had to index content[0].text) to the ballerinax/ai.anthropic ModelProvider. The chat() method returns an ai:ChatAssistantMessage where content is already a plain string? there's no content array to index into. Now the abstraction handles that internally.

.github/workflows/orchestrate-connector-updates.yml
TharaniDJ: Done

connector_automator/modules/sanitizor/spec_updater.bal
```diff
@@ -59,14 +59,29 @@ function updateParameterDescriptionInSpec(map<json> paths, string location, stri
...
                                 json|error parametersResult = operationMap.get("parameters");
                                 if parametersResult is json[] {
-                                    json[] parameters = parametersResult;
+                                    foreach int i in 0 ..< parametersResult.length() {
+                                        json param = parametersResult[i];
+                                        if param is map<json> {
+                                            map<json> paramMap = <map<json>>param;
+                                            if paramMap.hasKey("name") && paramMap.get("name") == paramName {
+                                                paramMap["description"] = description;
+                                                return ();
```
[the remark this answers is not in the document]
TharaniDJ: Implemented. updateParameterDescriptionInSpec now parses an optional ,in=<value> suffix from the location bracket (e.g., parameters[name=userId,in=path]). Both the operation-level and path-level parameter loops now check both name and in when a value is provided, and fall back to name-only matching for existing locations that don't include in= (full backward compatibility).

connector_automator/main.bal
```diff
@@ -589,26 +608,270 @@ function runFullPipeline(string... args) returns error? {
...
+            io:println("   Phase 2: Removing old test files and regenerating fresh...");
+
+            utils:CommandResult cleanupResult = utils:executeCommand(
+                string `rm -rf tests modules/mock.server`,
+                clientPath,
+                quietMode
+            );
+            if cleanupResult.exitCode == 0 {
+                io:println("   ✓ Old tests removed");
+            }
```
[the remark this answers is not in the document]
TharaniDJ: Implemented. Replaced rm -rf tests modules/mock.server shell call with two file:remove(path, file:RECURSIVE) calls one for clientPath + "/tests" and one for clientPath + "/modules/mock.server". Each failure is logged as a warning rather than aborting (the directories may simply not exist). Added import ballerina/file to main.bal imports.

connector_automator/modules/client_regenerator/sort_ballerina_type.bal
```diff
@@ -0,0 +1,212 @@
...
+
+    int firstTypeLine = -1;
+    int lastTypeLine = -1;
+
+    int i = 0;
+    while i < lines.length() {
+        string line = lines[i];
+        string stripped = line.trim();
+
+        if regex:matches(stripped, "public\\s+(type|const)") {
```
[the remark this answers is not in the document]
TharaniDJ: Implemented. Updated sort_ballerina_type.bal regex from "public\\s+(type|const)" to "public\\s+(type|const)\\b.*" so full declaration lines like public type Foo record {| match. Updated sort_ballerina_client.bal regex to append .* to the function detection pattern for the same reason

connector_automator/modules/code_fixer/prompts.bal
```diff
@@ -111,3 +111,229 @@ DO NOT include:
...
+File: ${filePath}
+</CONTEXT>
+
+<COMPILATION_ERRORS>
+${errorContext}
+</COMPILATION_ERRORS>
+
+<${typeContext}>
+
+<CURRENT_CODE>
```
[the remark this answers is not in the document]
TharaniDJ: Implemented. Both occurrences of <${typeContext}> have been replaced with a proper <TYPE_CONTEXT>\n${typeContext}\n</TYPE_CONTEXT> block. The typeContext string is now the body of a well-formed enclosing tag, not the tag name itself.

connector_automator/modules/example_generator/execute.bal
```diff
@@ -303,3 +346,100 @@ function printUsage() {
...
+                exampleName = exampleName.substring(1);
+            }
+
+            string mainBalPath = entry.absPath + "/main.bal";
+            boolean hasMain = check file:test(mainBalPath, file:EXISTS);
+            if hasMain {
+                exampleNames.push(exampleName);
+            }
+        }
+    }
```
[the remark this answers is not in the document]
TharaniDJ: Implemented. Replaced entry.absPath.substring(examplesPath.length()) with a lastIndexOf("/") basename extraction. exampleName is now derived purely from entry.absPath (which is always absolute per file:readDir), so relative/absolute mismatches between examplesPath and entry.absPath no longer affect the result. Downstream usage of connectorPath + "/examples/" + exampleName remains consistent.

connector_automator/modules/sanitizor/execute.bal
```diff
@@ -459,10 +459,66 @@ function convertAlignedYamlToJson(string alignedSpecPath, boolean quietMode = fa
...
+                log:printInfo("Successfully converted YAML to JSON using Python",
+                        yamlPath = yamlAlignedSpec,
+                        jsonPath = jsonAlignedSpec);
+            }
+            return;
+        }
+        
+        // All methods failed
+        return error("Failed to parse YAML content: " + jsonData.message() + 
+            ". Fallback tools (yq, python) also failed or not available.");
```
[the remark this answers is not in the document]
TharaniDJ: Implemented. Removed the backtick-to-single-quote replacement entirely. Backticks are valid in YAML string scalars (common in OpenAPI descriptions with Markdown), so the replacement was silently corrupting spec content. All three paths now operate on the original yamlContent.

connector_automator/modules/sanitizor/execute.bal
```diff
@@ -459,10 +459,66 @@ function convertAlignedYamlToJson(string alignedSpecPath, boolean quietMode = fa
...
+                        jsonPath = jsonAlignedSpec);
+            }
+            return;
+        }
+        
+        // Try python-based conversion as another fallback
+        utils:CommandResult pythonResult = utils:executeCommand(
+            string `python3 -c "import yaml, json, sys; print(json.dumps(yaml.safe_load(open('${yamlAlignedSpec}')), indent=2))"`,
+            "."
+        );
```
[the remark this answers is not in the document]
TharaniDJ: Implemented. The yq path is now wrapped in single quotes with embedded single quotes escaped ('\\''), eliminating shell injection via yamlAlignedSpec. For the python fallback, the command now reads from stdin (sys.stdin) instead of opening a file by path — the yaml content is written to a safe temp file (yamlAlignedSpec + ".stdin_tmp") and redirected via < stdinFile, so yamlAlignedSpec is never interpolated into the shell command. The temp file is removed after use with best-effort cleanup.

connector_automator/modules/code_fixer/prompts.bal
```diff
@@ -111,3 +111,229 @@ DO NOT include:
...
+**THE CORRECT COMBINED SOLUTION**:
+When BOTH patterns appear in history, you MUST use:
+${tripleBacktick}ballerina
+// For rest fields that return anydata, ALWAYS combine member access WITH type cast:
+string? fieldValue = <string?>response["fieldName"];
+
+// Or use ensureType for non-optional:
+string fieldValue = check response["fieldName"].ensureType();
+${tripleBacktick}
+
```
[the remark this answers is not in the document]
TharaniDJ: Fixed. Replaced the fenced code block (${tripleBacktick}ballerina...${tripleBacktick}) in the fix history section with plain indented examples. The LLM now sees the solution as plain text, reducing the chance it echoes code fences in its response.

connector_automator/modules/client_regenerator/sort_ballerina_client.bal
```diff
@@ -0,0 +1,253 @@
...
+    while i < lines.length() {
+        string line = lines[i];
+
+        if regex:matches(line, "\\s*resource\\s+isolated\\s+function\\s+(get|post|put|delete|patch)") {
+            if firstMethodLine == -1 {
+                firstMethodLine = i;
+            }
+
+            string[] methodLines = [line];
+            int startLine = i;
```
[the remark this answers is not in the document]
TharaniDJ: Already fixed in a previous session (.* appended to the regex). No change.

connector_automator/modules/example_generator/execute.bal
```diff
@@ -303,3 +346,100 @@ function printUsage() {
...
+            string exampleName = entry.absPath.substring(examplesPath.length());
+            if exampleName.startsWith("/") {
+                exampleName = exampleName.substring(1);
+            }
+
+            string mainBalPath = entry.absPath + "/main.bal";
+            boolean hasMain = check file:test(mainBalPath, file:EXISTS);
+            if hasMain {
+                exampleNames.push(exampleName);
+            }
```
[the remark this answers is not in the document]
TharaniDJ: Already fixed in a previous session (basename via lastIndexOf). No change.

connector_automator/modules/sanitizor/execute.bal
```diff
@@ -459,10 +459,66 @@ function convertAlignedYamlToJson(string alignedSpecPath, boolean quietMode = fa
...
+        if !quietMode {
+            log:printWarn("Ballerina YAML parser failed, trying yq fallback", 
+                errorMsg = jsonData.message());
+        }
+        
+        // Try using yq to convert YAML to JSON
+        utils:CommandResult yqResult = utils:executeCommand(
+            string `yq -o=json '.' "${yamlAlignedSpec}"`,
+            "."
+        );
```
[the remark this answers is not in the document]
TharaniDJ: Fixed. Both utils:executeCommand fallback calls now pass quietMode as the third argument, keeping log output consistent with the rest of the sanitizor flow.

connector_automator/modules/client_regenerator/sort_ballerina_type.bal
```diff
@@ -0,0 +1,218 @@
...
+                        name: typeName,
+                        typeKind: typeKind,
+                        startLine: startLine,
+                        endLine: i
+                    });
+                    i += 1;
+                    break;
+                }
+                i += 1;
+            }
```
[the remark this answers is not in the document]
TharaniDJ: Fixed. Added if braceCount == 0 { finalize + continue } guard right after computing braceCount. Used a separate openBraces variable in the inner while loop to avoid Ballerina's flow-narrowing restriction (BCE4017).

.github/workflows/auto_generate_connector_template.yml
```diff
@@ -8,30 +8,45 @@ on:
         required: false
         type: string
       ballerina-version:
-        description: "Ballerina version to use"
```
lnash94, line 11: Why were these descriptions removed?

connector_automator/docs/spec.md
```diff
@@ -1,1024 +0,0 @@
-# Connector Automator - Technical Specification
```
lnash94, line 1: Shall we keep this spec.md and update it with the necessary details?

connector_automator/modules/client_regenerator/analyze_version_change.bal
```diff
@@ -0,0 +1,109 @@
...
+    string changeType;
+    string[] breakingChanges;
+    string[] newFeatures;
+    string[] bugFixes;
+    string summary;
+    string confidence;
+};
+
+function analyzeWithAnthropic(string gitDiff) returns AnalysisResult|error {
+    string apiKey = os:getEnv("ANTHROPIC_API_KEY");
```
lnash94: Shall we add a constant for "ANTHROPIC_API_KEY" and use it?

connector_automator/modules/client_regenerator/sort_ballerina_client.bal
```diff
@@ -0,0 +1,296 @@
...
+        if token == "remote" {
+            isRemote = true;
+            break;
+        }
+    }
+
+    foreach int i in 0 ..< tokens.length() {
+        if tokens[i] == "function" && i + 1 < tokens.length() {
+            string nameOrMethod = tokens[i + 1];
+            if nameOrMethod == "get" || nameOrMethod == "post" || nameOrMethod == "put" ||
```
lnash94: Shall we define these HTTP method strings as constants to improve maintainability?

connector_automator/modules/client_regenerator/sort_ballerina_client.bal
```diff
@@ -0,0 +1,296 @@
...
+                return nameOrMethod;
+            }
+            if isRemote {
+                string lower = nameOrMethod.toLowerAscii();
+                if lower.startsWith("get") { return "get"; }
+                if lower.startsWith("post") { return "post"; }
+                if lower.startsWith("put") { return "put"; }
+                if lower.startsWith("delete") { return "delete"; }
+                if lower.startsWith("patch") { return "patch"; }
+                return "remote";
```
lnash94: When we are trying to find the function names, it sometimes associates with a specific HTTP method. However, most of the time, it will refer to the operation ID in the OpenAPI specification. Could you please explain this a bit more? How do we use this value?

connector_automator/modules/client_regenerator/sort_ballerina_client.bal
```diff
@@ -0,0 +1,296 @@
+import ballerina/io;
```
lnash94, line 17: Shall we add license headers?

connector_automator/modules/client_regenerator/sort_ballerina_type.bal
```diff
@@ -0,0 +1,236 @@
+import ballerina/io;
```
lnash94, line 17: Add license headers and check other places as well?

connector_automator/modules/client_regenerator/update_changelog.bal
```diff
@@ -0,0 +1,223 @@
...
+    } else {
+        io:println("Creating new CHANGELOG.md");
+        string changelogTemplate = string `# Change Log
+
+This file contains all the notable changes done to the Ballerina connector through the releases.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+${newUnreleasedSection}`;
+
```
lnash94: Shall we format this file?

connector_automator/README.md
```diff
@@ -1,325 +0,0 @@
-# Connector Automator
```
lnash94, line 1: Shall we keep this README.md and update it accordingly

connector_automator/modules/client_regenerator/analyze_version_change.bal
TharaniDJ: Done. Added const string ANTHROPIC_API_KEY_ENV = "ANTHROPIC_API_KEY"; and replaced the inline string literal and the error message to use it consistently.

connector_automator/modules/client_regenerator/sort_ballerina_client.bal
TharaniDJ: Done. Added seven module-level constants (METHOD_GET, METHOD_POST, METHOD_PUT, METHOD_DELETE, METHOD_PATCH, METHOD_REMOTE, METHOD_UNKNOWN) and replaced all inline string literals throughout extractMethodType, extractPath, and generateSortKey with the constants.

connector_automator/modules/client_regenerator/sort_ballerina_client.bal
TharaniDJ: Ballerina connectors can be generated with either resource functions (e.g. resource function get /users[string id]()) or remote functions (e.g. remote function getUserById(string id)). For resource functions, the HTTP verb is a keyword in the declaration, so it's extracted directly. For remote functions there is no HTTP verb keyword ,only the operation ID from the OpenAPI spec is used as the function name. We infer the HTTP method from the operation ID prefix (get, list → GET; create, post → POST; etc.) so that remote functions can still be sorted by their semantic HTTP method alongside resource functions. If no known prefix matches, the function falls back to the remote priority bucket (sorted after all HTTP method based functions). This is purely a sorting heuristic and does not affect how the function is called.

.github/workflows/orchestrate-connector-updates.yml
TharaniDJ: Removed this part.

connector_automator/docs/spec.md
TharaniDJ: Sorry, I updated this file

connector_automator/modules/client_regenerator/sort_ballerina_client.bal
TharaniDJ: I improved this part

connector_automator/README.md
TharaniDJ: Done

connector_automator/modules/client_regenerator/sort_ballerina_type.bal
TharaniDJ: Ok, done

connector_automator/modules/client_regenerator/sort_ballerina_client.bal
TharaniDJ: Ok, Done

connector_automator/docs/spec.md
TharaniDJ: Done

.github/workflows/auto_generate_connector_template.yml
TharaniDJ: Sorry, again added

.github/workflows/auto_generate_connector_template.yml
```diff
@@ -200,17 +248,17 @@ jobs:
...
             bal version
           fi
 
-          # create Config.toml for the connector automator. 
+          # create Config.toml for the connector automator.
 
           cat > Config.toml << EOF
           [connector_automator.code_fixer]
-          maxIterations = 3
+          maxIterations = 5
```
ThisaruGuruge: Can we define this as a variable on the top? Would be okay even if we define it as an input. 
The reason is this value is buried deep inside the file, so can't notice it right away.

.github/workflows/auto_generate_connector_template.yml
```diff
@@ -130,7 +149,7 @@ jobs:
       - name: Clone Connector Automator
         run: |
           echo "Cloning connector automator tool from ballerina-library..."
-           git clone -b connector-automation https://github.com/nostoc/ballerina-library.git temp-library
+           git clone -b main https://github.com/ballerina-platform/ballerina-library.git temp-library
```
ThisaruGuruge: Instead of cloning directly to the repo root, shall we use `$RUNNER_TEMP/` to clone these additional files? This way, we don't have to clean them up separately, and also don't need to bother about cleaning up before git operations.

.github/workflows/auto_generate_connector_template.yml
```diff
@@ -156,24 +181,46 @@ jobs:
       - name: Read File Extension
```
ThisaruGuruge, line 189: Aren't we reading the file extension in the previous step?

.github/workflows/auto_generate_connector_template.yml
```diff
@@ -156,24 +181,46 @@ jobs:
...
           fi
+
+          # Get the single file
+          if [ -f "openapi.json" ]; then
+            file="openapi.json"
+          elif [ -f "openapi.yaml" ]; then
+            file="openapi.yaml"
+          elif [ -f "openapi.yml" ]; then
+            file="openapi.yml"
+          fi
```
ThisaruGuruge: Since we identify the file in the above step, do we need this if-else block again?

.github/workflows/orchestrate-connector-updates.yml
```diff
@@ -0,0 +1,487 @@
...
+            sleep 30
+          fi
+
+      - name: Collect all connector PR information
+        if: steps.trigger_connectors.outputs.triggered_count > 0
+        id: collect_prs
+        env:
+          GH_TOKEN: ${{ secrets.BALLERINA_BOT_TOKEN }}
+        run: |
+          echo "Collecting PR information from all triggered connectors..."
```
ThisaruGuruge: For longer scripts like this, would be better to keep them as separate script files and call them from the workflow. Let's check whether that works (might have to consider how we can handle this when running from a connector repo, but since we are anyway cloning the library repo, would be possible. This is a minor detail, so if it takes a lot of effort, we can park this.

.github/workflows/orchestrate-connector-updates.yml
```diff
@@ -0,0 +1,487 @@
...
+          </html>
+          EOF
+
+          echo "Generated email body"
+
+      - name: Check email secrets
+        if: steps.trigger_connectors.outputs.triggered_count > 0
+        id: check_email_secrets
+        run: |
+          if [ -z "${{ secrets.GMAIL_USERNAME }}" ] || [ -z "${{ secrets.GMAIL_APP_PASSWORD }}" ]; then
```
ThisaruGuruge: Let's use `EMAIL_USERENAME` and `EMAIL_PASSWORD`.

.github/workflows/orchestrate-connector-updates.yml
```diff
@@ -0,0 +1,487 @@
...
+          else
+            echo "Email secrets are configured"
+            echo "email_configured=true" >> $GITHUB_OUTPUT
+          fi
+
+      - name: Send consolidated email
+        if: steps.trigger_connectors.outputs.triggered_count > 0 && steps.check_email_secrets.outputs.email_configured == 'true'
+        continue-on-error: true
+        id: send_email
+        uses: dawidd6/action-send-mail@v3
```
ThisaruGuruge: We can use a [Ballerina script to send emails](https://ballerina.io/learn/by-example/send-email/). I am -1 for using 3rd party GitHub actions since this action also requires using secrets.

connector_automator/docs/spec.md
```diff
@@ -114,13 +116,24 @@ The `runFullPipeline` function orchestrates all modules:
...
 }
 ```
 
+The pipeline accepts the following arguments:
+
+| Position | Argument | Description |
+|----------|----------|-------------|
+| 0 | `specPath` | Path to the OpenAPI specification file |
+| 1 | `outputPath` | Root directory for generated output |
+| 2 | `autoYes` | `"yes"` to skip confirmation prompts |
```
ThisaruGuruge: `autoYes` seems an odd name to have.

connector_automator/docs/spec.md
```diff
@@ -298,6 +360,69 @@ bal openapi -i <spec> --mode client -o <output> \
...
+**Sort order** (by `methodPriority` map):
+
+| Priority | HTTP Method / Type |
+|----------|--------------------|
+| 1 | `get` |
+| 2 | `post` |
+| 3 | `put` |
+| 4 | `patch` |
+| 5 | `delete` |
+| 6 | `remote` (non-resource remote functions) |
```
ThisaruGuruge:
```suggestion
| 6 | `remote` |
```

connector_automator/docs/spec.md
```diff
@@ -547,16 +672,60 @@ AI-powered automatic resolution of Ballerina compilation errors.
...
+   f. Record this attempt in fileFixHistory[filePath]
+   g. If user confirms (or autoYes):
       - Create backup
       - Apply fix
 6. If no fixes applied → Stop
 7. Repeat from step 1 (max iterations)
 8. Final build check and summary
 ```
 
+**Configurable**: `maxIterations` is read from `Config.toml` (`[connector_automator.code_fixer] maxIterations = 5`).
```
ThisaruGuruge, line 688: Is this applicable? I saw a hard-coded `5` as the number of iterations.

connector_automator/modules/client_regenerator/analyze_version_change.bal
```diff
@@ -0,0 +1,242 @@
...
+        lastErr = response;
+        if attempt < MAX_RETRIES {
+            io:println(string `API call failed (attempt ${attempt}/${MAX_RETRIES}): ${response.message()}. Retrying in ${delay}s...`);
+            error? sleepErr = runtime:sleep(delay);
+            if sleepErr is error {
+                io:println(string `Sleep interrupted: ${sleepErr.message()}`);
+            }
+            delay *= 2.0d;
+        }
+    }
```
ThisaruGuruge: Instead of implementing the retry logic manually, we can just pass the retry configs to the client initialization. (Check the comment in the `buildModel()` method.

connector_automator/modules/client_regenerator/analyze_version_change.bal
```diff
@@ -0,0 +1,242 @@
...
+function buildModel() returns ai:ModelProvider|error {
+    string apiKey = os:getEnv(ANTHROPIC_API_KEY_ENV);
+    if apiKey == "" {
+        return error(string `${ANTHROPIC_API_KEY_ENV} environment variable is not set`);
+    }
+    return check new anthropic:ModelProvider(
+        apiKey,
+        anthropic:CLAUDE_SONNET_4_6,
+        maxTokens = 4096
+    );
```
ThisaruGuruge, line 78: Add retry config here:


```suggestion
    return check new anthropic:ModelProvider(
        apiKey,
        anthropic:CLAUDE_SONNET_4_6,
        maxTokens = 4096,
        retryConfig = {
            count: MAX_RETRIES,
            interval: RETRY_INITIAL_DELAY,
            backOffFactor: BACK_OFF_FACTOR, // Define a variable or a constant for this
            maxWaitInterval: ...
        }
    );
```

This way, you don't need the `chatWithRetries` method, you can just call `model->chat()` directly. Retry will be handled by the client.

connector_automator/modules/client_regenerator/sort_ballerina_client.bal
```diff
@@ -0,0 +1,331 @@
...
+        }
+    }
+
+    // For resource functions: HTTP method is the keyword immediately before the path
+    // e.g. `resource function get /users[string id](...)`
+    foreach int i in 0 ..< tokens.length() {
+        if tokens[i] == "function" && i + 1 < tokens.length() {
+            string nameOrMethod = tokens[i + 1];
+            if nameOrMethod == METHOD_GET || nameOrMethod == METHOD_POST || nameOrMethod == METHOD_PUT ||
+                    nameOrMethod == METHOD_DELETE || nameOrMethod == METHOD_PATCH {
```
ThisaruGuruge: If we define methods as an enum, this condition would be simplified to:

```suggestion
            if nameOrMethod is HttpMethod {
```

connector_automator/modules/client_regenerator/sort_ballerina_client.bal
```diff
@@ -0,0 +1,331 @@
...
+    return METHOD_UNKNOWN;
+}
+
+function extractHttpMethodFromBody(string content) returns string {
+    string lower = content.toLowerAscii();
+    if lower.includes("->get(") { return METHOD_GET; }
+    if lower.includes("->post(") { return METHOD_POST; }
+    if lower.includes("->put(") { return METHOD_PUT; }
+    if lower.includes("->delete(") { return METHOD_DELETE; }
+    if lower.includes("->patch(") { return METHOD_PATCH; }
```
ThisaruGuruge: Would be better to define a map for this case:

```
map<string> methodMap = {
    "->get(": METHOD_GET,
    "->post(": METHOD_POST,
    //...
};
```

Then this function would be 

```ballerina
if methodMap.hasKey(lower) {
    return methodMap.get(lower);
}
return METHOD_REMOTE;
```

connector_automator/modules/client_regenerator/sort_ballerina_client.bal
```diff
@@ -0,0 +1,331 @@
...
+                string[] nameParts = regex:split(funcName, "\\(");
+                return nameParts.length() > 0 ? nameParts[0] : funcName;
+            }
+            break;
+        }
+    }
+
+    foreach int i in 0 ..< tokens.length() {
+        if (tokens[i] == METHOD_GET || tokens[i] == METHOD_POST || tokens[i] == METHOD_PUT ||
+                tokens[i] == METHOD_DELETE || tokens[i] == METHOD_PATCH) && i + 1 < tokens.length() {
```
ThisaruGuruge:
```suggestion
        if tokens[i] is HttpMethod && i + 1 < tokens.length() {
```

connector_automator/modules/client_regenerator/sort_ballerina_client.bal
```diff
@@ -0,0 +1,331 @@
...
+
+import ballerina/io;
+import ballerina/file;
+import ballerina/regex;
+
+const string METHOD_GET = "get";
+const string METHOD_POST = "post";
+const string METHOD_PUT = "put";
+const string METHOD_DELETE = "delete";
+const string METHOD_PATCH = "patch";
```
ThisaruGuruge: Can define these as an enum:

```suggestion
enum HttpMethod {
    METHOD_GET = "get",
    METHOD_POST = "post",
    METHOD_PUT = "put",
    METHOD_DELETE = "delete",
    METHOD_PATCH = "patch"
}
```

connector_automator/modules/client_regenerator/sort_ballerina_client.bal
```diff
@@ -0,0 +1,331 @@
...
+        string path = extractPath(block.content);
+        [string, string, string] sortKey = generateSortKey(methodType, path);
+
+        methods.push({
+            content: block.content,
+            startLine: block.startLine,
+            endLine: block.endLine,
+            methodType: methodType,
+            path: path,
+            sortKey: sortKey
```
ThisaruGuruge: Can use shorthand notation where the field name and the variable name is the same.

```suggestion
            methodType,
            path,
            sortKey
```

connector_automator/modules/client_regenerator/sort_ballerina_type.bal
```diff
@@ -0,0 +1,252 @@
...
+            if braceCount == 0 {
+                lastTypeLine = startLine;
+                string typeContent = line;
+                string typeKind = extractTypeKind(typeContent);
+                typeDefs.push({
+                    content: typeContent,
+                    name: typeName,
+                    typeKind: typeKind,
+                    startLine: startLine,
+                    endLine: startLine
```
ThisaruGuruge: start and end lines are the same. Is this expected?

connector_automator/modules/client_regenerator/sort_ballerina_type.bal
```diff
@@ -0,0 +1,252 @@
...
+                    name: typeName,
+                    typeKind: typeKind,
+                    startLine: startLine,
+                    endLine: i
+                });
+                i += 1;
+                continue;
+            }
+
+            int braceCount = countChar(line, "{") - countChar(line, "}");
```
ThisaruGuruge: Would be better to define constant for open and close braces since we are using them in many places.

connector_automator/modules/client_regenerator/sort_ballerina_type.bal
```diff
@@ -0,0 +1,252 @@
...
+
+function extractAllTypes(string content) returns [TypeDefinition[], int, int] {
+    string[] lines = regex:split(content, "\n");
+    TypeDefinition[] typeDefs = [];
+
+    int firstTypeLine = -1;
+    int lastTypeLine = -1;
+
+    int i = 0;
+    while i < lines.length() {
```
ThisaruGuruge, line 120: This while loop seems too long. Consider breaking it into smaller function calls and early returns/breaks.

connector_automator/modules/code_fixer/code_fixer.bal
```diff
@@ -84,8 +84,157 @@ function prepareErrorContext(CompilationError[] errors) returns string {
...
+            }
+        }
+    }
+
+    return typeNames;
+}
+
+// Check if a string is a Ballerina keyword
+function isBalKeyword(string s) returns boolean {
+    string[] keywords = ["error", "nil", "string", "int", "boolean", "float", "decimal", "anydata", "json", "byte", "any"];
```
ThisaruGuruge, line 118: What's the purpose of checking `isBalKeyword`? This list is not exhaustive and we will definitely miss some keywords.

connector_automator/modules/code_fixer/code_fixer.bal
```diff
@@ -84,8 +84,157 @@ function prepareErrorContext(CompilationError[] errors) returns string {
...
+        // Try mock.server types.bal if it's a mock file error
+        if filePath.includes("mock") {
+            typesFilePath = projectPath + "/modules/mock.server/types.bal";
+            typesContent = io:fileReadString(typesFilePath);
+        }
+    }
+
+    // Try to read client.bal for additional context on method signatures
+    string clientFilePath = projectPath + "/client.bal";
+    string|io:Error clientContent = io:fileReadString(clientFilePath);
```
ThisaruGuruge, line 159: Looks like we are ignoring the error here. Is it intentional? In that case, put a comment explaining why we are ignoring the error.

connector_automator/modules/code_fixer/code_fixer.bal
```diff
@@ -136,6 +294,41 @@ public function fixFileWithLLM(string projectPath, string filePath, CompilationE
     };
 }
 
+// Build a context string from previous fix attempts
+function buildFixHistoryContext(FixAttempt[] attempts) returns string {
```
ThisaruGuruge, line 302: Nice!

connector_automator/main.bal
```diff
@@ -589,26 +637,300 @@ function runFullPipeline(string... args) returns error? {
...
+            printStepHeader(4, "Regenerating Tests for New API Version", quietMode);
+            string[] testArgs = [outputDir, sanitizedSpec];
+            testArgs.push(...pipelineOptions);
+            error? testResult = test_generator:executeTestGen(...testArgs);
+            if testResult is error {
+                io:println(string `✗ Test regeneration failed: ${testResult.message()}`);
+                io:println("   Cannot continue without valid tests");
+                return testResult;
+            }
+            io:println("✓ Test regeneration completed successfully");
```
ThisaruGuruge: These long if-else ladders is a code smell. Consider breaking it into smaller functions.
Soemthing like:

```ballerina
if !utils:hasCompilationErrors(phase1BuildResult) {
    continueToExamples(); //... Just pseudo code. Define expressive methods and call them here.
} else {
    regenerateFresh();
}
```

connector_automator/modules/sanitizor/spec_updater.bal
```diff
@@ -59,17 +66,36 @@ function updateParameterDescriptionInSpec(map<json> paths, string location, stri
...
+                                            boolean nameMatches = paramMap.hasKey("name") && paramMap.get("name") == paramName;
+                                            boolean inMatches = paramIn == "" || (paramMap.hasKey("in") && paramMap.get("in") == paramIn);
+                                            if nameMatches && inMatches {
+                                                paramMap["description"] = description;
+                                                return ();
+                                            }
+                                        }
+                                    }
+                                }
+                            }
```
ThisaruGuruge: Code smell–higher [cognitive complexity](https://www.sonarsource.com/blog/5-clean-code-tips-for-reducing-cognitive-complexity#tip-3-nesting-things-can-make-things-bad-really-fast). Let's try to reduce this.

.github/workflows/auto_generate_connector_template.yml
TharaniDJ: Added it as a max-iterations workflow input (default: 5) so it's visible at the top of the file.

.github/workflows/auto_generate_connector_template.yml
TharaniDJ: Yes, when openapi-url is provided the extension is already known from the URL. Set EXTENSION and SPEC_FILE in the download step directly. Restricted "Read File Extension" to only run when openapi-url == '' (spec already committed to the repo).

.github/workflows/auto_generate_connector_template.yml
TharaniDJ: Removed

.github/workflows/orchestrate-connector-updates.yml
TharaniDJ: I implemented this change

connector_automator/docs/spec.md
TharaniDJ: Renamed to skipConfirmation — clearer that it suppresses interactive prompts rather than implying a yes/no boolean flag.

connector_automator/docs/spec.md
TharaniDJ: Fixed in the previous set of changes. MaxIterations is now exposed as a max-iterations workflow input (default: 5) and injected into Config.toml via ${{ inputs.max-iterations }}

connector_automator/modules/client_regenerator/analyze_version_change.bal
TharaniDJ: Removed chatWithRetry entirely and passed retryConfig directly to anthropic:ModelProvider. Added BACK_OFF_FACTOR and MAX_WAIT_INTERVAL constants alongside the existing ones. The model->chat() call is now plain and retry is handled by the client.

connector_automator/modules/client_regenerator/sort_ballerina_type.bal
TharaniDJ: Intentional when the declaration header has no net-open brace (braceCount == 0) the entire type sits on a single line, so startLine == endLine. Added an explanatory comment.

connector_automator/modules/code_fixer/code_fixer.bal
TharaniDJ: Added a comment explaining that the list is intentionally minimal: a missed keyword causes a harmless extra search, not a failure.

connector_automator/modules/code_fixer/code_fixer.bal
TharaniDJ: Added a comment: client.bal is optional context , if it doesn't exist, the fixer proceeds with types.bal alone.

connector_automator/modules/code_fixer/code_fixer.bal
TharaniDJ: Thank you

ThisaruGuruge (approved):

PR #9131, opened 2026-09-07
Split the status dashboard by distribution packaging

niveathika:
## Purpose

The `Ballerina Modules` table described itself as "published under the `ballerina` organization and packed with the Ballerina distribution", conflating two facts that [ballerina-spec#1462](https://github.com/ballerina-platform/ballerina-spec/issues/1462) separates: which organization a module is published under, and whether it ships inside the installer.

Diffing `stdlib_modules.json` against `ballerina-distribution`'s `gradle.properties` shows four `ballerina`-org modules are already Central-only — `zip`, `pdf`, `xlsx` and `smb` — so the table has been inaccurate for all four.

## Approach

`library_modules` entries take an optional `packed_with_distribution` flag (defaults to `true`), and the generator renders the array as two tables: `Ballerina Modules` and `Ballerina Central-Only Modules`.

Keeping one array rather than adding a second one means `release/main.bal` (`release_libs`) and `trigger-graalvm-check.bal` keep consuming `library_modules` unchanged, so the Central-only modules are still released and GraalVM-checked. The flag is written to `stdlib_modules.json` for library modules only — stamping `packed_with_distribution: true` on 113 `ballerinax` connectors, drivers and tools would be plainly false.

The `Level` column is now derived from each module's own level rather than a counter seeded at `1`, which would have labelled the Central-only table's first row `1` regardless of its real level. Levels stay global across the whole graph, so a four-row subset legitimately shows gaps (1, 2, 4, 5).

The `Ballerina Extended Modules` blurb also said "not packed with the Ballerina distribution", which would now read as the same thing as the new section; it says `ballerinax` organization instead.

## Two pre-existing gaps fixed along the way

- `ai.np` had its version key inferred as `stdlibAi.npVersion`, which never matched the distribution's `stdlibAiNpVersion`, so its dependency edges were never resolved. Now set explicitly.
- `module-ballerina-mcp` ships in the distribution but was absent from `module_list.json` entirely, so it never appeared on the dashboard.

Both were missing from the dependency graph, so levels shift: `mcp` lands at level 8, pushing `ai` 8 → 9 and `ai.np` 9 → 10.

## Checklist

- [x] Generator run end-to-end against the GitHub API; output reproducible across two runs
- [x] `bal test` in `dashboard/` — 4 passing
- [x] Linked an issue

The README diff is 40 lines: the four moved rows, the new section, and ten rows whose Trivy badge flipped to `disabled` — current state the scheduled workflow would have picked up anyway.

- Split the status dashboard into `Ballerina Modules` and `Ballerina Central-Only Modules` tables.
- Track Central-only modules in `module_list.json` and partition dashboard data by module name.
- Derive dashboard module levels from dependency graph data.
- Add `module-ballerina-mcp`, correct the `module-ballerina-ai.np` version key, and update dependency levels.
- Update module packaging and release workflow documentation.
- Add dashboard coverage for Central-only module partitioning.


docs/build-and-release-guide.md
```diff
@@ -180,7 +180,7 @@ The Ballerina Library Release Manager will use [Library Release](https://github.
 7. Once the module list is updated, notify the team to refrain from merging any PRs to the default branches, including the post-release PRs.
 8. Then the release manager can trigger the [Library Release Workflow](https://github.com/ballerina-platform/ballerina-library/actions/workflows/release_pipeline.yml) to trigger the release.
     This workflow expects the following inputs:
-    - `release_libs`: To release the library (`module-ballerina-*` modules that are packed with the distribution) modules. These modules are listed under the `modules` field in the `module_list.json` file. The default value is `true`.
+    - `release_libs`: To release the library (`module-ballerina-*`) modules, whether or not they are packed with the distribution. These modules are listed under the `library_modules` field in the `module_list.json` file. The default value is `true`.
```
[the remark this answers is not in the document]
niveathika: Fixed in https://github.com/ballerina-platform/ballerina-library/commit/2089ddd — `release_libs` now points to `stdlib_modules.json`, which is what `release/main.bal` actually reads. The other four bullets in this section (`release_extensions`, `release_tools`, `release_handwritten_connectors`, `release_generated_connectors`) have the same wrong filename, but that's pre-existing and outside this PR's diff, so left as-is here.

docs/adding-a-new-ballerina-module.md
```diff
@@ -399,7 +399,7 @@ Add the module to the relevant array based on the type of the module.
 
 >**Note:** Do not edit the `stdlib_modules.json` file manually. It will be auto-generated once the `module_list.json` file is updated.
 
-The JSON entry has two fields.
+The JSON entry has three fields.
```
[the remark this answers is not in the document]
niveathika: Fair point — fixed in https://github.com/ballerina-platform/ballerina-library/commit/2089ddd by dropping the count ("The JSON entry supports the following fields.") instead of claiming a fixed number of fields.

docs/adding-a-new-ballerina-module.md
```diff
@@ -389,17 +389,21 @@ This step is required only if the module is a part of the Ballerina library and
 
 To add a module to the Ballerina library, add an entry in the [`module_list.json`](https://github.com/ballerina-platform/ballerina-library/blob/main/release/resources/module_list.json) file in the [`ballerina-library`](https://github.com/ballerina-platform/ballerina-library) repository. This JSON file has 4 main fields.
```
ThisaruGuruge: This is a preexisting bug; but let's fix it:
```suggestion
To add a module to the Ballerina library, add an entry in the [`module_list.json`](https://github.com/ballerina-platform/ballerina-library/blob/main/release/resources/module_list.json) file in the [`ballerina-library`](https://github.com/ballerina-platform/ballerina-library) repository. This JSON file has 5 main fields.
```

dashboard/main.bal
```diff
@@ -298,20 +303,31 @@ function updateDashboard(List moduleDetails) returns error? {
 }
 
 isolated function getLibraryModulesDashboard(Module[] modules) returns string|error {
```
ThisaruGuruge, line 317: Shall we add a simple test to check this functionality?

docs/adding-a-new-ballerina-module.md
niveathika: Fixed in https://github.com/ballerina-platform/ballerina-library/commit/00c692e — updated the count to 5.

dashboard/main.bal
niveathika: Added in https://github.com/ballerina-platform/ballerina-library/commit/00c692e — extracted the split into `partitionByDistributionPackaging` and added `partitionByDistributionPackagingTest`, which checks all four Central-only modules and `module-ballerina-io` land in the correct partition.

ThisaruGuruge (dismissed):
LGTM

Better if we omit the README.md changes so the next workflow trigger will correct it. Otherwise we will have to resolve conflicts everyday.

PR #9155, opened 2026-09-11
Add bal scan job to the security scan workflow

daneshk:
## Summary
- Adds a `bal-scan` job (runs `bal tool pull scan` + `bal scan`) alongside the existing Trivy job, and renames the workflow to **Security Scan**.
- The job fails only on `VULNERABILITY` findings; `CODE_SMELL` findings are uploaded as a report artifact but don't fail the job, since existing repos carry pre-existing code smells (e.g. `module-ballerina-http` alone currently has 494) that shouldn't break nightly runs across the org.
- The active workflow logic moves to `security-scan-template.yml`. `trivy-scan-template.yml` becomes a thin forwarding stub (`uses: ./.github/workflows/security-scan-template.yml`) so the ~70 repos still referencing it by that exact path keep working unchanged. Callers can migrate to `security-scan-template.yml` directly over time; the stub can be removed once everyone has.
- Adds `security-scan.yml` to the `generated-connector-template` scaffold (under `library-templates/`) so newly generated connector repos get the security scan workflow wired up automatically, pointing directly at `security-scan-template.yml`.

## Verification
- Ran `bal scan` locally against `module-ballerina-http` and `module-ballerina-cache` to confirm: report location (`target/report/scan_results.json`), that `bal scan` always exits 0 regardless of findings (so pass/fail must be derived from the report), and the `rule.ruleKind` field (`VULNERABILITY` vs `CODE_SMELL`) used for the fail gate.
- Confirmed `bal scan` needs the package's native jar dependencies already built (per `Ballerina.toml` platform dependencies), so the job runs the same `./gradlew build -x test` step used by the Trivy job before scanning.
- Validated all three YAML files parse correctly.

## Test plan
- [ ] Merge and confirm the `Security Scan` workflow runs successfully end-to-end on `ballerina-library`'s own next scheduled trigger (via a caller repo)
- [ ] Spot-check one existing repo still calling `trivy-scan-template.yml@main` to confirm the forwarding stub runs both jobs without changes on their side
- [ ] Generate a new connector repo via `apply-library-repo-templates.yml` and confirm `security-scan.yml` is created and resolves correctly

Related: wso2-enterprise/integration-engineering#2865

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Adds a reusable **Security Scan** workflow for Ballerina packages, examples, and optional scan paths.

The workflow creates separate reports, uploads them to code scanning, and fails when findings require remediation. Code-quality findings do not fail the job.

Retains the previous Trivy workflow as a forwarding stub. Adds the workflow to the generated connector template with manual and scheduled triggers. Improves input handling, action pinning, build artifact creation, and credential handling.

### DIFF
--- a/.github/workflows/security-scan-template.yml
+++ b/.github/workflows/security-scan-template.yml
@@ -14,43 +14,8 @@ on:
         default: "ballerina"
 
 jobs:
-  trivy-scan:
-    name: Trivy Vulnerability Scan
-    runs-on: ubuntu-22.04
-    if: github.repository_owner == 'ballerina-platform'
-    steps:
-      - uses: actions/checkout@v3
-      - name: Set up JDK 21
-        uses: actions/setup-java@v3
-        with:
-          distribution: "temurin"
-          java-version: 21.0.3
-
-      - name: Build with Gradle
-        env:
-          packageUser: ${{ github.actor }}
-          packagePAT: ${{ secrets.GITHUB_TOKEN }}
-        run: ./gradlew build -x test ${{ inputs.additional-build-flags }}
-
-      - name: Create lib directory if not exists
-        run: mkdir -p ballerina/lib
-
-      - name: Run Trivy vulnerability scanner
-        uses: aquasecurity/trivy-action@master
-        env:
-          TRIVY_DB_REPOSITORY: ghcr.io/aquasecurity/trivy-db,public.ecr.aws/aquasecurity/trivy-db
-          TRIVY_JAVA_DB_REPOSITORY: ghcr.io/aquasecurity/trivy-java-db,public.ecr.aws/aquasecurity/trivy-java-db
-        with:
-          scan-type: "rootfs"
-          scan-ref: "${{ github.workspace }}/ballerina/lib"
-          format: "table"
-          timeout: "10m0s"
-          exit-code: "1"
-          scanners: "vuln"
-          cache-dir: "/tmp/trivy-cache"
-
-  bal-scan:
-    name: Ballerina Static Code Analysis
+  security-scan:
+    name: Security Scan
     runs-on: ubuntu-22.04
     if: github.repository_owner == 'ballerina-platform'
     steps:
@@ -75,14 +40,36 @@ jobs:
           distribution: "temurin"
           java-version: 21.0.3
 
-      # bal scan needs the package to compile, which for Java-interop modules means the
-      # native jar(s) referenced under [[platform.javaXX.dependency]] in Ballerina.toml
-      # must already be built.
+      # Built once and shared by both scans below: Trivy scans the built jars under
+      # ballerina/lib, and bal scan needs the native jar(s) referenced under
+      # [[platform.javaXX.dependency]] in Ballerina.toml to already be built in order
+      # to compile the package.
       - name: Build with Gradle
         env:
           packageUser: ${{ github.actor }}
           packagePAT: ${{ secrets.GITHUB_TOKEN }}
         run: ./gradlew build -x test ${{ inputs.additional-build-flags }}
+
+      - name: Create lib directory if not exists
+        run: mkdir -p ballerina/lib
+
+      # continue-on-error so a Trivy finding doesn't skip the bal scan steps below;
+      # its outcome is checked explicitly in the final step instead.
+      - name: Run Trivy vulnerability scanner
+        id: trivy
+        continue-on-error: true
+        uses: aquasecurity/trivy-action@master
+        env:
+          TRIVY_DB_REPOSITORY: ghcr.io/aquasecurity/trivy-db,public.ecr.aws/aquasecurity/trivy-db
+          TRIVY_JAVA_DB_REPOSITORY: ghcr.io/aquasecurity/trivy-java-db,public.ecr.aws/aquasecurity/trivy-java-db
+        with:
+          scan-type: "rootfs"
+          scan-ref: "${{ github.workspace }}/ballerina/lib"
+          format: "table"
+          timeout: "10m0s"
+          exit-code: "1"
+          scanners: "vuln"
+          cache-dir: "/tmp/trivy-cache"
 
       - name: Pull the bal scan tool
         run: bal tool pull scan
@@ -102,15 +89,26 @@ jobs:
 
       # bal scan always exits 0, so pass/fail is derived from the report: real
       # VULNERABILITY findings fail the job, CODE_SMELL findings are reported only
-      # (via the uploaded artifact) since existing style debt shouldn't break nightly runs.
+      # (via the uploaded artifact) since existing style debt shouldn't break nightly
+      # runs. Combined here with the Trivy outcome so the job fails if either scanner
+      # found something, regardless of order.
       - name: Fail the job if vulnerabilities are found
         working-directory: ${{ inputs.scan-path }}
         run: |
           VULN_COUNT=$(jq '[.[] | select(.rule.ruleKind == "VULNERABILITY")] | length' target/report/scan_results.json)
           SMELL_COUNT=$(jq '[.[] | select(.rule.ruleKind == "CODE_SMELL")] | length' target/report/scan_results.json)
           echo "bal scan found $VULN_COUNT vulnerability finding(s) and $SMELL_COUNT code smell finding(s). See the bal-scan-report artifact for details."
+
+          FAIL=0
           if [ "$VULN_COUNT" -gt 0 ]; then
             jq '[.[] | select(.rule.ruleKind == "VULNERABILITY")]' target/report/scan_results.json
-            echo "::error::bal scan found $VULN_COUNT vulnerability finding(s). Failing the job."
+            echo "::error::bal scan found $VULN_COUNT vulnerability finding(s)."
+            FAIL=1
+          fi
+          if [ "${{ steps.trivy.outcome }}" == "failure" ]; then
+            echo "::error::Trivy vulnerability scanner found issue(s). See the 'Run Trivy vulnerability scanner' step above."
+            FAIL=1
+          fi
+          if [ "$FAIL" -eq 1 ]; then
             exit 1
           fi
--- a/.github/workflows/security-scan-template.yml
+++ b/.github/workflows/security-scan-template.yml
@@ -1,0 +1,116 @@
+name: Security Scan
+
+on:
+  workflow_call:
+    inputs:
+      additional-build-flags:
+        required: false
+        type: string
+        default: ""
+      scan-path:
+        description: "Path to the Ballerina package to run 'bal scan' against"
+        required: false
+        type: string
+        default: "ballerina"
+
+jobs:
+  trivy-scan:
+    name: Trivy Vulnerability Scan
+    runs-on: ubuntu-22.04
+    if: github.repository_owner == 'ballerina-platform'
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up JDK 21
+        uses: actions/setup-java@v3
+        with:
+          distribution: "temurin"
+          java-version: 21.0.3
+
+      - name: Build with Gradle
+        env:
+          packageUser: ${{ github.actor }}
+          packagePAT: ${{ secrets.GITHUB_TOKEN }}
+        run: ./gradlew build -x test ${{ inputs.additional-build-flags }}
+
+      - name: Create lib directory if not exists
+        run: mkdir -p ballerina/lib
+
+      - name: Run Trivy vulnerability scanner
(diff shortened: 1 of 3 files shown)


daneshk:
Security Scan workflow: https://github.com/ballerina-platform/module-ballerina-cache/actions/runs/34992540918

daneshk:
Addressed all 3 CodeRabbit findings in d7a1a9f:

1. **Shell injection via unquoted workflow inputs** (L33, L156): `additional-build-flags` and `scan-path` are now passed through `env:` and read as shell variables (a quoted bash array for the Gradle flags, `--arg` for jq) instead of being spliced directly into the `run:` script text. Verified locally that a `-x test; echo pwned` payload is now inert data rather than a second executed command.
2. **Mutable `@master` ref for a third-party action** (L70): pinned `aquasecurity/trivy-action` to the commit SHA for its `v0.36.0` release.
3. **Unnecessary `secrets: inherit`** in the generated-connector-template scaffold caller: removed it — the called workflow only reads `secrets.GITHUB_TOKEN`, which GitHub grants to called reusable workflows automatically.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

.github/workflows/security-scan-template.yml
```diff
@@ -0,0 +1,195 @@
...
+        type: string
+        default: ""
+      scan-path:
+        description: "Path to the Ballerina package to run 'bal scan' against"
+        required: false
+        type: string
+        default: "ballerina"
+
+jobs:
+  build:
```
TharmiganK, line 22: Why do we need a separate build workflow?

.github/workflows/security-scan-template.yml
```diff
@@ -0,0 +1,195 @@
...
+            (.runs[].results[].locations[].physicalLocation.artifactLocation.uri) |= ($prefix + .)
+          ' target/report/scan_results.sarif > target/report/scan_results_github.sarif
+
+      # Publishes findings to the repo's Security > Code scanning tab, where they're
+      # tracked and auto-closed across runs. Non-blocking: some caller repos may not
+      # grant security-events: write to the token (permissions here are a ceiling, not
+      # a guarantee), and that shouldn't fail the scan itself -- only the gate below does.
+      - name: Upload SARIF to GitHub code scanning
+        continue-on-error: true
+        uses: github/codeql-action/upload-sarif@v3
```
TharmiganK: Seeing this warning:

```
Warning: CodeQL Action v3 will be deprecated in December 2026. Please update all occurrences of the CodeQL Action in your workflow files to v4. For more information, see https://github.blog/changelog/2025-10-28-upcoming-deprecation-of-codeql-action-v3/
```

.github/workflows/security-scan-template.yml
```diff
@@ -0,0 +1,195 @@
...
+
+      - name: Pull the bal scan tool
+        run: bal tool pull scan
+
+      # --format=sarif and the default (JSON) report are mutually exclusive per
+      # invocation, so we standardize on SARIF alone: it's what both the code
+      # scanning upload and the vulnerability gate below need, and --scan-report
+      # still gives us the HTML report in the same run.
+      - name: Run bal scan
+        working-directory: ${{ inputs.scan-path }}
```
TharmiganK: We should support multiple paths right? We also have examples

.github/workflows/trivy-scan-template.yml
```diff
@@ -9,40 +13,8 @@ on:
...
-        with:
-          scan-type: "rootfs"
-          scan-ref: "${{ github.workspace }}/ballerina/lib"
-          format: "table"
-          timeout: "10m0s"
-          exit-code: "1"
-          scanners: "vuln"
-          cache-dir: "/tmp/trivy-cache"
+  forward:
+    uses: ./.github/workflows/security-scan-template.yml
```
TharmiganK, line 17: With this the Trivy scan will also run bal scan in the library repos right? Are we going to change the library workflow naming and the status bar we add to the README.md files

.github/workflows/security-scan-template.yml
```diff
@@ -0,0 +1,195 @@
...
+          retention-days: 1
+
+  trivy-scan:
+    name: Trivy Vulnerability Scan
+    needs: build
+    runs-on: ubuntu-22.04
+    if: github.repository_owner == 'ballerina-platform'
+    steps:
+      - name: Download build output
+        uses: actions/download-artifact@v4
```
YasanPunch: Is it okay that this job no longer checks out the repo? Trivy reads `.trivyignore` from the working directory by default, and the old template ran after `actions/checkout`, so callers such as module-ballerinax-salesforce that suppress a CVE that way might be affected?

.github/workflows/security-scan-template.yml
```diff
@@ -0,0 +1,195 @@
...
+      # bal scan needs to resolve the platform dependencies declared in Ballerina.toml
+      # in order to compile the package.
+      - name: Upload build output
+        uses: actions/upload-artifact@v4
+        with:
+          name: build-output
+          path: |
+            ballerina/lib/**
+            **/build/libs/*.jar
+          if-no-files-found: warn
```
YasanPunch, line 78: What happens for a module with no `ballerina/lib` and no `build/libs` jars, e.g. a pure Ballerina connector? It seems upload-artifact v4 creates no artifact when nothing matches, and download-artifact v4 then fails with `Artifact not found`, so both scan jobs might fail before the `mkdir -p` step runs?

YasanPunch:
Small thing on the description: the Verification section mentions `scan_results.json` and `rule.ruleKind`, but the workflow gates on `level == "warning"` in `scan_results.sarif`. Maybe update the text to match?

.github/workflows/security-scan-template.yml
daneshk: trivy scan and bal scan both needs build artifacts to do the scanning. Doing bal build for each scan is not efficient. Hence I separate the bal build and reuse the build artifacts for both scaning workflow.

.github/workflows/trivy-scan-template.yml
daneshk: yes, we gradually needs to move all repos to use the scan workflow. I will send the PRs for those separately. still we get them merge, we use the trivy scan flow to check the security check

.github/workflows/security-scan-template.yml
daneshk: Confirmed this is real — checked locally and 11 repos have a repo-root `.trivyignore` (e.g. `module-ballerinax-salesforce` suppresses `CVE-2024-6763`). Fixed in 37aa857: added `actions/checkout` back to `trivy-scan`.

.github/workflows/security-scan-template.yml
daneshk: Confirmed real — checked module-ballerinax-aws.dynamodb, which has no native subproject and no externalJars, matches zero files for both upload patterns, and already calls this template today. Fixed in 37aa857 by always touching a placeholder file into ballerina/lib before upload, so upload-artifact never skips creating the artifact.

.github/workflows/security-scan-template.yml
daneshk: Fixed in 37aa857: bumped github/codeql-action/upload-sarif from v3 to v4 (confirmed identical inputs, safe drop-in).

.github/workflows/security-scan-template.yml
daneshk: Added in 37aa857 + 2b71684: bal scan now runs against the main package, every examples/*/ subdirectory that is its own Ballerina package (examples/ itself usually isn't a package -- it's a folder of them, sometimes nested two levels deep like examples/snowpeak/service), and any caller-supplied additional-scan-paths (comma-separated input). Verified locally end-to-end against module-ballerina-http's 7 packages.

Also split library vs examples into separate artifacts and code-scanning categories (2b71684), since a vulnerability in the library is a direct risk to consumers while one in an example is about our own reference code quality -- worth tracking as distinct signals.

daneshk:
@coderabbitai please recheck

.github/workflows/security-scan-template.yml
```diff
@@ -0,0 +1,318 @@
...
+      # Guarantees at least one file always matches below: a module with no native
+      # subproject and no externalJars (e.g. module-ballerinax-aws.dynamodb) produces
+      # zero files for both patterns, and upload-artifact creates no artifact at all
+      # when nothing matches (if-no-files-found: warn only suppresses the failure, not
+      # the missing artifact) -- which would then fail download-artifact in both scan
+      # jobs with "Artifact not found".
+      - name: Ensure build output is never empty
+        run: |
+          mkdir -p ballerina/lib
+          touch ballerina/lib/.keep
```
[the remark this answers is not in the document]
daneshk: Good catch, confirmed real: verified via upload-artifact's own action.yml that include-hidden-files defaults to 'false', so `.keep` would have been silently excluded, defeating the whole point of the fix. Fixed in aa0c37e — renamed to `ballerina/lib/artifact-placeholder`.

.github/workflows/security-scan-template.yml
```diff
@@ -0,0 +1,318 @@
...
+    name: Trivy Vulnerability Scan
+    needs: build
+    runs-on: ubuntu-22.04
+    if: github.repository_owner == 'ballerina-platform'
+    steps:
+      # Needed so a repo-root .trivyignore (used by several repos, e.g.
+      # module-ballerinax-salesforce, to suppress a specific already-triaged CVE) is
+      # present -- Trivy reads it from the working directory by default. Without this,
+      # previously-suppressed CVEs would resurface as new failures.
+      - uses: actions/checkout@v3
```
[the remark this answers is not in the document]
daneshk: Fixed in aa0c37e:
- Bumped actions/checkout v3 -> v7.0.1, actions/setup-java v3 -> v6.0.1 (both deprecated upstream).
- Pinned every remaining mutable GitHub-maintained action to a commit SHA: checkout, setup-java, upload-artifact (also bumped to v7.0.1), download-artifact (bumped to v8.0.1, confirmed input-compatible), and upload-sarif (confirmed v4 is still current, no v5 exists).
- Added persist-credentials: false to all three checkout steps per the zizmor artipacked warning -- none of these jobs push or commit.

Deliberately left ballerina-platform/ballerina-library/.github/actions/resolve-java-version@main and ballerina-platform/setup-ballerina@v1.1.3 unpinned: the former is a same-repo, actively-evolving shared utility that the org's own convention (PR #9159) already references via @main across multiple templates; the latter already targets a specific numbered release, matching every other template in this repo.

daneshk:
Addressed CodeRabbit's latest round (aa0c37e, 03e6531) and re-tested live against two real repos:

**Fixed:**
1. **Critical**: `ballerina/lib/.keep` (the empty-artifact placeholder) is a dotfile, and `upload-artifact@v4` excludes hidden files by default — so the earlier fix didn't actually fix anything. Renamed to `artifact-placeholder`.
2. Bumped/pinned every remaining mutable action reference to a commit SHA: `checkout` (v3→v7.0.1), `setup-java` (v3→v6.0.1), `upload-artifact` (v4→v7.0.1), `download-artifact` (v4→v8.0.1, confirmed input-compatible), `upload-sarif` (confirmed v4 is current). Added `persist-credentials: false` to all checkout steps (zizmor's artipacked warning).
3. Deliberately left `ballerina-platform/ballerina-library/.github/actions/resolve-java-version@main` and `ballerina-platform/setup-ballerina@v1.1.3` unpinned — same-repo/org convention, explained in the review reply.

**Found a new bug via live testing** (not from static review — only showed up running against a repo with multiple examples): GitHub's code scanning stopped accepting multiple SARIF runs under one category (changed July 2025). `module-ballerina-http` has 6 example packages, so the `bal-scan-examples` upload silently produced zero alerts despite reporting "success" (continue-on-error masked it). Fixed by merging each group's per-package SARIF into one combined run before upload.

**Live-tested against 2 real repos after all fixes:**
- `module-ballerinax-aws.dynamodb` (zero native jars, zero externalJars — the case the empty-artifact fix targets): all 3 jobs passed, artifact correctly created with 1 file (the placeholder), 3 packages discovered (library + 2 examples), 0 findings.
- `module-ballerina-http` (6 example packages): Build and Trivy passed; `bal-scan` correctly failed only at the final gate — `Library: 0 vulnerability finding(s)`, `Examples: 28 vulnerability finding(s)` (real, pre-existing issues like permissive CORS in `examples/snowpeak`, not caused by this PR). Confirmed both `bal-scan-library` and `bal-scan-examples` now show up as separate Security tab analyses with correct alert locations.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

